PR #163, opened 2019-02-20
Enable battery compute module for generic system

nickdiorio:
Enable battery for Generic system within startup script.


sjanzou:
Fixed with patch branch commit in ssc 7cb4deae595c01ae40ed2c01646a5432c49ef9cb

PR #168, opened 2019-04-04
Rad cool GUI changes

adyreson:
This adds inputs to the GUI for CSP radiative cooling. This goes with ssc pull request #219. Right now, the inputs are part of the Rankine Cycle page. The key variable is the condenser type (Radiative in the drop down or CT=4). On 1-8-19 I met with Ty and Nick and we discussed separating into a new cycle type "Rankine Cycle with Radiative Cooling and Cold Storage", but for now the selection of the condenser type is used to enable the Cold Storage Type (ctes_type) if Radiative condenser is selected. If it is not selected, then ctes_type is disabled. This should avoid conflicts that would arise from selecting air-cooled, evaporative, or hybrid and also selecting cold storage.  Ideally, other variables would also be conditionally enabled according to the following logic: If Condenser type = CT=4= Rad cool with cold storage, then: Allow choice of Turbine inlet pressure control to include the two new cycles I've added: 'Sliding pressure for high backpressure' and 'Sliding pressure for low backpressure' in addition to the other two choices. 
If Storage type = 2, then: Enable new radiative cooling with cold storage parameters. 
If Storage type >2, then: Enable new radiative cooling with cold storage parameters except, these, which should be disabled off: 'Percent of htf in warm tank initially', 'Minimum allowable tank height`.


nickdiorio:
@adyreson , I think the clahub check is failing.  If you click through it, everything should pass. We're just testing CLA hub out.

PR #177, opened 2019-05-01
PySAM

dguittet:
Added api_autogen and api directories with targets export_config and libSAM_API, respectively. These are used for making the SAM library. In the future, a streamlined build for libSAM_api will be implemented as currently there are some manual steps involved.

From runtime/defaults removed hard-coded paths to resource files and replaced with empty strings. This is necessary to have clean exported defaults as json files which are then used in PySAM.

SAM GUI, export config & libSAM_api building on VS2017 & Mac. Setting up travis-ci to use CMake builds


dguittet:
reopening as PySAM2 branch

PR #179, opened 2019-05-03
Interconnection

nickdiorio:
This pull request adds:

- The "Grid" page to the following models:
1. pvsamv1
2. pvwatts
3. fuelcell
4. wind
5. geothermal
6. generic

It could be added to any of the other models as well.  Defaults set to 100 MW grid interconnection, so a user will not inadvertently run up against a limit within explicitly changing it (or creating a giant system).  We could make it more sophisticated by adding a checkbox or something that enables the interconnection, but for now this seems adequate.

Benefits of this approach is that any changes we add to "grid" page and underlying model instantly gets pulled in across all different technologies using it.

### DIFF
--- a/deploy/runtime/lossdiag.lk
+++ b/deploy/runtime/lossdiag.lk
@@ -51,6 +51,12 @@ loss_diagram{'Flat Plate PV'} = define() {
 			add_loss_term( 'annual_ac_lifetime_loss_percent', 'AC Lifetime daily losses- year one' );
 	add_loss_term('annual_xfmr_loss_percent', 'Transformer loss' );	
 	add_loss_term('annual_ac_perf_adj_loss_percent', 'AC Availability and curtailment' );	
+	if (varinfo('enable_interconnection_limit')){
+		if (value('enable_interconnection_limit')){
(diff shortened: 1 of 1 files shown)


janinefreeman:
@nickdiorio I built and ran it through the UI, seems to behave about like I'd expect. I would advocate for adding an enable/disable checkbox, rather than just having users set the interconnect limit to be really high. That is more consistent with what we've done historically, and I think will be clearer for the user.

janinefreeman:
Sorry, a couple of other quick notes:
-I don't see a total annual loss (kWh or %) due to interconnection limits? I think those values need to be added.
-The loss also needs to be added to the loss tree for the PV model.
-Looks like the battery doesn't dispatch with the grid interconnection limit in mind? If not, maybe worth adding a git issue to that effect. I think that would be an important addition.

Thanks!

nickdiorio:
@janinefreeman , I added:

- checkbox to enable or disable interconnection limit
- annual loss reporting due to interconnection limit
- updated the loss diagram
- created an issue for battery dispatch https://github.com/NREL/SAM/issues/181

Thanks for the review!  I'll go ahead and merge.

PR #238, opened 2020-01-24
Solar Resource Data and Battery UI fixes

cpaulgilman:
Clean up Solar Resource Data UI callback.

Fix battery environment temperature checks.

Set default battery environment temperature option to single value for all configurations to avoid UI warning for Generic System configurations.


dguittet:
I think it makes sense to make the default a fixed environment temp as most people will be doing some type of temp control!

I'm not sure what the solar resource callback changes are doing, what can I do to test them?

Otherwise I think it's good to merge!

PR #258, opened 2020-03-17
Inputs Browser - update to handle inf and NaN values

sjanzou:
Address issue when comparing inf and NaN values to other numbers and each other
Before update
image

After update
image
image

### DIFF
--- a/src/variables.cpp
+++ b/src/variables.cpp
@@ -890,11 +890,11 @@ bool VarValue::ValueEqual( VarValue &rhs )
 							double x = m_val(r, c);
 							double y = rhs.m_val(r, c);
 							if ((std::isnan(x)) || (std::isnan(y)))
-								equal = equal && ((std::isnan(x)) && (std::isnan(y)));
+								equal = ((std::isnan(x)) && (std::isnan(y)));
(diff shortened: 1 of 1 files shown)


src/variables.cpp
```diff
@@ -886,8 +886,16 @@ bool VarValue::ValueEqual( VarValue &rhs )
...
+						{
+							double x = m_val(r, c);
+							double y = rhs.m_val(r, c);
+							if ((std::isnan(x)) || (std::isnan(y)))
+								equal = equal && ((std::isnan(x)) && (std::isnan(y)));
+							else if ((std::isinf(x)) || (std::isinf(y)))
+								equal = equal && ((std::isinf(x)) && (std::isinf(y)));
+							else
+								equal = equal && (RelDif(m_val(r, c), rhs.m_val(r, c)) < TOLERANCE);
+						}
```
dguittet: This makes more sense, nice!

I think `equal = equal && <>` can be replaced with simply `equal = <>` since we're in an if-statement ensuring that equal = 1 and `1 && x = x`. 

Then after `else` if not equal, you can break to cut the iteration short.

PR #273, opened 2020-04-01
LandBOSSE Cost Model Integration

### LINKED ISSUE #270: Integrate Wind LandBOSSE via Python to SAM
Adding the new Land-based Balance-of-System Systems Engineering [LandBOSSE](https://github.com/WISDEM/LandBOSSE) model v2.2.7, a package in python, to be used in windpower simulations in SAM. 

This will allow estimating the balance of system costs part of `total_installed_cost` for utility scale plants, and will be enabled for all PPA and the Merchant Plant configurations for Wind.

This project introduces the addition of a Python interpreter into SAM, which will allow future integration with other Python packages. The Python and LandBOSSE installation is optional and triggered via an enable toggle.

Docs to add:
 - Windows: How to update via pip the version of LandBOSSE

dguittet:
**Features**
In PPA configurations of Wind, add LandBOSSE v2.2.7.3 cost model in System Costs page.

-  Installation triggered by "enable" button
-  Can be disabled by same button from simulation
-  Run within equations, parametrics & regular simulation

**Code changes**
in deploy/runtime/python:

 - scripts for installing Python or Miniconda by SAM: install_python.ps1 (Windows) and install_python.sh (Unix) used by SAM executable to install Python to deploy/runtime/python
 - python_config.json: records where Python and pip executables are & what packages are installed
 - landbosse.json: records version, read by SAM for correct installation

source files:

 - pythonhandler.h/cpp for SAM to load & write python configurations, and run python installer & pip processes
 - invoke.cpp: new functions for installing & running landbosse in the GUI
 - variables.h/cpp: new Merge function for VarTables
 - main.h/.cpp: functions for installing Python & using pip

defaults & ui:

 - Modified PPA configs of windpower to use LandBOSSE cost model

CmakeLists.txt:

 - fix Linux warnings & build

close #270

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,10 @@ before_script:
   - export WEXDIR=/home/travis/build/$user/wex
   - export SSCDIR=/home/travis/build/$user/ssc
   - export SAMNTDIR=/home/travis/build/$user/SAM
-  - git clone https://github.com/google/googletest /home/travis/build/$user/googletest
-  - git clone https://github.com/$user/lk $LKDIR
-  - git clone https://github.com/$user/wex $WEXDIR
-  - git clone https://github.com/$user/ssc $SSCDIR
+  - git clone --depth=1 https://github.com/google/googletest /home/travis/build/$user/googletest
+  - git clone --depth=1 https://github.com/$user/lk $LKDIR
+  - git clone --depth=1 https://github.com/$user/wex $WEXDIR
+  - git clone --depth=1 https://github.com/$user/ssc $SSCDIR
   
 script:
   - export PATH=$PATH:$HOME/wx-3.1.1/bin
--- a/deploy/runtime/ui/Battery Bank Sizing.txt
+++ b/deploy/runtime/ui/Battery Bank Sizing.txt
@@ -872,7 +872,8 @@ batt_bank_power_dc_ac
 Battery desired bank power DC or AC
  
 Battery Sizing
-0
+5
+DC|AC
 1
 1
 1
@@ -900,7 +901,8 @@ batt_bank_size_dc_ac
 Battery desired bank capacity DC or AC
  
 Battery Sizing
-0
+5
+DC|AC
 1
 1
 1
--- a/deploy/runtime/ui/Battery Dispatch Front of Meter.txt
+++ b/deploy/runtime/ui/Battery Dispatch Front of Meter.txt
@@ -1166,7 +1166,7 @@ BatteryDispatch
 1
 1
 0
-Default
+DataArray
 batt_cycle_cost
 3
 1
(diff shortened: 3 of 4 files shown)


brtietz:
I'm getting a few issues building on Windows. The most critical one seems to be:

Severity	Code	Description	Project	File	Line	Suppression State
Error	C1083	Cannot open include file: 'AtlBase.h': No such file or directory	SAM	C:\Users\bmirletz\source\repos\sam_dev\sam\src\pythonhandler.cpp	6	

This is after checking out the landbosse-integration branches on both SAM and SSC, and re-running cmake in the build folder.

Additionally, should the README be updated to reference JSON and Python as included libraries?

dguittet:
Thanks for testing it out!

> Severity Code Description Project File Line Suppression State
> Error C1083 Cannot open include file: 'AtlBase.h': No such file or directory SAM C:\Users\bmirletz\source\repos\sam_dev\sam\src\pythonhandler.cpp 6

It looks like some C++ development tools for Visual Studio need to be installed: https://github.com/microsoft/terminal/issues/483

And yes, thanks for the reminder to update the README & About.

deploy/runtime/python/install_python.ps1
```diff
@@ -0,0 +1,143 @@
...
+    try {
+        Invoke-WebRequest -OutFile $filename $url
+    }
+    catch {
+        throw "failed to download $url"
+    }
+}
+
+
+function Get-Python {
```
daniel-thom, line 42: This function has two behavioral inconsistencies with the Linux/Mac install script.
1. It does not support the FORCE_DOWNLOAD env variable to allow skipping the download for devs. This is minor, but we should make it consistent.
2. The script will fail if the to-be-extracted Python directory already exists.  The Miniconda install command now includes `-u' to update whereas it used to fail. I forgot to make a comparable change here.  Just need to call `Remove-Item` on the directory if it exists.

daniel-thom:
Do we want to add support for upgrading the LandBOSSE version as part of this feature?  A possible implementation is this:
- A SAM patch would change the LandBOSSE version in the config file.
- SAM/SSC would detect a version mismatch at a TBD point in time and then run pip to upgrade.
- Question:  How should we handle the case where the user has moved the SAM directory after LandBOSSE was installed?  In this scenario pip will not work.  We would need to either re-run the Python install script or ask the user to reinstall SAM.

deploy/runtime/python/install_python.ps1
dguittet: Yes, good points. Would you please add those options?

dguittet:
Regarding the moving of the SAM directory, I'm actually not getting an error. In my case, python_config.json is referring to pip directly rather than a symlink to pip (sometimes `pip` is a symlink to `pip3`). I'm seeing the same non-error behavior on Windows now linking to the `pip.exe` executable.

dguittet:
minimal simulation result changes in all configs:
image

PR #277, opened 2020-04-08
Fix one form sizing error in patch

sjanzou:


sjanzou:
Updated default results and one form sizing change made and tested for patch 2 and merging into develop

PR #290, opened 2020-04-27
Fix macros

cpaulgilman:
Combine Cases:

* Use Year 1 data for techs that run lifetime simulation. In future, may want to combine lifetime data, but this solution is good for now.

* Include battery replacement cost in operating costs for combined system.

* For no financial and third party / host, set installation and O&M cost to zero so they can be used in a combined system

* Use macro input AC degradation, not degradation inputs from cases.

PV Sizing:

* FIx table cell coloring.

* Module list no longer includes modules that exceed inverter Vmppt_max

### DIFF
--- a/deploy/runtime/defaults/Generic System_All Equity Partnership Flip.txt
+++ b/deploy/runtime/defaults/Generic System_All Equity Partnership Flip.txt
@@ -1,5 +1,5 @@
 1
-335
+334
 adjust
 1
 5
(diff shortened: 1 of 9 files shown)


janinefreeman:
Very slick improvements to the combine cases macro! I like the Notes feature.

The PV sizing macro hangs SAM for me on the default detailed/Single Owner case- if it's not doing that for you, maybe Brian could test on his machine?

cpaulgilman:
I can replicate that...investigating.

janinefreeman:
Looks great, Paul! Everything seems to be working for me now. One note that the "generic system/SO" defaults on this branch appear to be out of date- not sure if patch needs to be merged in, or if they need to be updated in patch. Thanks!!

PR #297, opened 2020-05-04
Travis runs test_script_ow.lk

brtietz:
Changes:

- Allow SAM to run lk scripts from the command line
- Bring in CMakeLists changes required for Linux build into patch. (when merging with develop, resolve conflicts to develop)
- Add Dockerfile that allows us to simulate the Travis environment locally
- Add lines to travis script that runs test_script_ow.lk at the end of the build
- Add python script that compares results csv generated in travis to the one committed to the branch

These tests ran as expected in https://travis-ci.org/github/NREL/SAM/builds/683091139 The build failed due to the changes to test_results_linux64 over the last two months, and the python script enumerated the changes and failed the build. test_results_linux64 will be updated in another PR.

The monster line in the travis script (line 32 for Linux) does the following:

- Starts test_script_ow.lk in the background
- Every minute, prints out the number of completed tests (based on the number of lines in test_results_linux64)
- When the number of lines in travis' test_results_linux64 matches the one committed to the branch, proceed to the python script to compare lines

I have not yet tested the osx version of the script (because the Linux script is still failing). Once the new linux results get merged into patch, we can look for a passing run on both operating systems. @dguittet please point out anything that is obviously wrong in the osx paths.

### DIFF
--- a/test_results_linux64.csv
+++ b/test_results_linux64.csv
@@ -114,7 +114,6 @@ Solar Water Heating,Residential,2252.006104,35.191990,NA,;
 Solar Water Heating,Commercial,2252.080811,23.101711,NA,; 
 Solar Water Heating,LCOE Calculator,2281.620361,NA,NA,; 
 Solar Water Heating,None,2281.620361,NA,NA,; 
-Biopower,Commercial,353982848,9.413036,NA,; 
 Biopower,Single Owner,353982848,13.339031,13.681313,; 
 Biopower,Merchant Plant,353982848,12.177007,NA,; 
 Biopower,Leveraged Partnership Flip,353982848,13.083217,13.587696,; 


janinefreeman:
Thanks for tackling this! Looking forward to discussing the procedure for using this functionality on Thursday (which I think means including an updated "test_results_linux64" with any commit that should change the answers?). Quick question- what will happen if the number of lines in travis' test_results exceeds the number of lines in the one committed to the branch?

brtietz:
Correct on the procedure. OSX will also need to be updated.

Adding a line to the actual results could pass the tests, if all of the expected results were present and matched. The python script loads two dictionaries based on the test results, and as long as everything in the expected results is present in the actual results, the python script will pass it.

However, if the new test showed up in the middle, then an expected result (the last one) would be missing in the actual results, and the test would fail. The current setup has a minor race condition that way.

dguittet:
Regarding the case with different number of lines, how should new configurations be handled? New configurations being tested will produce new lines that may be in the middle.

brtietz:
> 
> 
> Regarding the case with different number of lines, how should new configurations be handled? New configurations being tested will produce new lines that may be in the middle.

Ultimately, we need to update the expected results when there are new configurations. For builds in the meantime, I think we have two choices:

(1) Let the build pass, since all of the expected results are present. The order of the tests doesn't matter, since the python script is comparing them based on name, not line order in the file. Right now the last test takes less than 60 seconds so odds are all tests would be present in the actual results and the build would pass. The worst case here is that the last test would fail, which might be confusing.

(2) Fail the build if there is a mismatch in the number of entries. The python script could check the number of entries in the csv, and fail if they don't match.

Preferences?

dguittet:
Say I added new configurations in a feature branch & updated the expected results in the csv on that feature branch. When I merge into develop, say, I think it passes as long as the existing config results are the same and my new configs aren't NaN.

My question is just that if you initiate your comparison python script based on number of lines in the output file, how will it know there haven't been new, potentially long-running configurations added?

brtietz:
> My question is just that if you initiate your comparison python script based on number of lines in the output file, how will it know there haven't been new, potentially long-running configurations added?

Currently, it won't. But if they're not in the expected results there's nothing to check against. If they're in the middle, the build will fail since the last configurations in the expected results weren't tested. If they're at the end, the build will pass.

Is that ok, or should the Python script check the number of lines?

brtietz:
Or are you saying the build should be able to pass in the case where len(actual_results) > len(expected_results)?

dguittet:
I think the fact that the build will fail if the new configurations are in the middle is probably not okay. This is a relatively rare occurrence, but in the last year we've added a new technology configuration (MHK) and modified the number of financial models (removed some) and split out batteries into its own configuration. 

So I do think some way to account for passing that would be more accurate. Perhaps if there's some end of file marker produced so that you can start the comparison only once it's finished.

brtietz:
Ah, ok, I hadn't thought of the number of cases decreasing. In that case the intermediate build would still fail, since there are expected results that aren't present, but failing that in 55 mins is better than 120.

Code for that is being tested in the latest push.

janinefreeman:
If I understand this correctly, wouldn't adding/removing configs be the same procedure as updating an LCOE for a changed default? So if you add a new technology, for example, you'd be in charge of updating the test results file to reflect those new configs. The person reviewing your PR would have to confirm that those changes make sense, then Travis should be building the code that corresponds with the changes you've made to the test results file and checking against that-- much the same way that we can change test answers in SSC, and someone has to review that those changes make sense. We don't want the only commits that pass to be unchanged results in the defaults file, because those should change sometimes. Unless I'm missing some subtlety?

brtietz:
Yes, that's how it _should_ work. The python script will always be checking against whatever defaults are committed to the branch, which ideally have been updated appropriately. Darice and I have been discussing the edge cases for when someone forgets to updates the defaults. 

I'm expecting a human will always update the defaults file, I get nervous about automated updates.

dguittet:
I think maybe I misunderstood the intended workflow. What's the "correct" CSV?

In the case when I'm merging a feature branch into develop, is it supposed to be testing whether the simulation results match the feature branch CSV, or whether the results do not change drastically from the CSV on develop?

dguittet:
@sjanzou to the convo

dguittet:
Okay, I see that we're testing that the simulation results are the same as the dev-updated CSV, which we should be treating as the answer template. 

So if the line lengths of the file are different, then it should be a fail since that means the dev forgot to update the CSV.

During merges, it's up to the dev to compare that the CSV changes are appropriate.

Sorry for my confusing remarks then!

janinefreeman:
That was my thought process on intended workflow, but we can discuss further Thursday if there's a better or more complete desired workflow!

sjanzou:
When the defaults are not updated and the script overwrites the defaults so the simulation can run, the values used to overwrite the default values are taken as the "default" values that were entered in the UI Interface editor on the right-hand side under "Default Value"... As Brian noted, this adds a line to the csv results file. So, as Darice mentioned, this is a failure. thanks for adding me to the conversation.

brtietz:
Current test failures are due to battery defaults changes in: https://github.com/NREL/SAM/pull/304 and https://github.com/NREL/SAM/pull/303 (plus one CSP failure, not sure what's happening there). Should I update the Windows defaults in this branch to get a pass prior to merging?

dguittet:
I'm sorry I had forgotten to update the test results due to the changes to the batt_temp & batt_resistance. Could you please update those?

brtietz:
Looks like Steve updated them, I just had to pull them into this branch. I'll update the Linux results as well shortly.

brtietz:
Merged to patch in https://github.com/NREL/SAM/commit/6856cc25fdbaaeafdfd90005bccacc76f86f6db1

PR #299, opened 2020-05-05
update linux test results based on the command line scripts branch

brtietz:
I couldn't get the Linux build to run just on patch (given the CMakeLists changes) so I ran it on Command_line_script. Please make sure these are the expected results based on changes since March.


brtietz:
Build passes with latest push: https://travis-ci.org/github/NREL/SAM/builds/685741708 this merges to Command_line_script, so it should be safe to merge.

PR #305, opened 2020-05-13
Sam 285

sjanzou:
Merged into patch but dod not want to merge into develop until Darice and Paul review


sjanzou:
test_results_win64.csv are from patch branch and should not be merged into develop.

cpaulgilman:
Thanks for this fix, Steve -- looks good to me. I updated all of the DataArray variables I could find and added some notes as a UI callback comment to the AA Widgets.txt form for documentation. Other objects with a mode property that may need a similar fix are DataLifetimeArray and DataLifetimeMatrix.

dguittet:
The variable descriptions look wonderful!

image

Issue #16, opened 2017-08-31
Electricity bill calculation improvements


cpaulgilman:
* Apply end-of-year credit to following year for California NEM: https://sam.nrel.gov/forum/forum-general/1159

* Change month for end-of-year kWh rollover credit
For Residential and Commercial financial models, SAM applies any kWh rollover credits remaining in December 31 to the December electricity bill. Some states specify a different month, for example Maryland is end of April. Request is to add an input to specify the month for the "end of year" credit: https://sam.nrel.gov/forum/forum-general/1824. Duke Energy NC "resets accumulated excess on May 31" https://sam.nrel.gov/forum/forum-general/2170

* Allow separate TOU schedules for Saturday and Sunday per NamPower rates in Namibia: https://www.nampower.com.na/Page.aspx?p=178, and also as typical in South Africa: https://sam.nrel.gov/forum/forum-general/2969

brtietz:
"Change month for end-of-year kWh rollover credit" is merged in to develop.

"Apply end-of-year credit to following year for California NEM" is partially implemented, the current implementation effectively delays the check by a month. See comments in https://github.com/NREL/ssc/pull/382 for how to improve the model.

Would recommend either implementing those comments or removing the checkbox prior to the 2021 release.

sjanzou:
@brtietz, is the UI list of months enabled for all metering options?

brtietz:
Yes

PR #324, opened 2020-06-02
Excess gen credit options

brtietz:
Implements two out of three enhancements in NREL/SAM#16. Adds a drop down menu to the utility rate GUI page to allow users to select the month, as well as a checkbox if the kWh credits should be applied to the following month. The checkbox is disabled for configurations besides net metering. Add defaults for these new variables to all behind the meter configurations.

Pairs with SSC PR: https://github.com/NREL/ssc/pull/382


cpaulgilman:
See my comments for NREL/ssc#382.

PR #331, opened 2020-06-12
Fix some if-else expressions where indenting makes them unclear

bje-:
This silences G++ warnings about same.

The first hunk adds braces because, by my reading of the code, this is what was intended (and the indenting seems to strongly suggest it, too). The remaining hunks seem to be correct, but need the indenting corrected.

### DIFF
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -764,9 +764,9 @@ void fcall_value( lk::invoke_t &cxt )
 			    bool trigger = true;
 			    if (cxt.arg_count() == 3 ) {
 			        trigger = cxt.arg(2).as_boolean();
-				if (trigger){
-				    cc.GetCase().VariableChanged( name );
-				}
+			    }
+			    if (trigger) {
+			      cc.GetCase().VariableChanged( name );
 			    }
 			}
 			else


src/invoke.cpp
```diff
@@ -762,11 +762,12 @@ void fcall_value( lk::invoke_t &cxt )
 		{
 			if ( vv->Read( cxt.arg(1), false ) ){
 			    bool trigger = true;
-			    if (cxt.arg_count() == 3 )
+			    if (cxt.arg_count() == 3 ) {
```
janinefreeman, line 765: While I agree that this seems to be what is implied by the indentation, adding this curly brace will fundamentally change the behavior of this function. @sjanzou , can you comment whether this looks like the intended behavior, or whether we just need to change indenting for clarity?

sjanzou: @janinefreeman, good catch, the proposed change is incorrect (verified in VS2019) 
image
and needs to be 
image


and needs to be:

bje-:
Thanks a lot. It was certainly hard to tell which way the bracing was meant to go. Fixed.

bje-:
Ping?

PR #332, opened 2020-06-14
Eliminate use of fgets() and fixed sized buffers

bje-:
Let's use C++ so we don't have to use fixed sized buffers with all of their associated problems!


src/main.cpp
```diff
@@ -2629,12 +2629,14 @@ int SamApp::RevisionNumber()
 #else
 	wxString path = SamApp::GetRuntimePath() + "/patches/patch_lnx.txt";
 #endif
-	if ( FILE *fp = fopen( (const char*)path.c_str(), "r" ) )
+	std::ifstream fp;
```
sjanzou, line 2632: Changes would have to be tested with Version Upgrade script and with web updated

src/widgets.cpp
```diff
@@ -2310,8 +2312,9 @@ class AFDataLifetimeArrayDialog : public wxDialog
 		{
 			wxFileDialog dlg(this, "Select data file to import");
 			if (dlg.ShowModal() != wxID_OK) return;
-			FILE *fp = fopen(dlg.GetPath().c_str(), "r");
-			if (!fp)
+			std::ifstream fp;
```
sjanzou, line 2315: Tested by @dguittet - should be fine

src/invoke.cpp
```diff
@@ -3447,19 +3448,20 @@ static void fcall_sam_async( lk::invoke_t &cxt )
 	cxt.result().empty_hash();
 
 	lk_string fn = cxt.arg(0).as_string();
-	FILE *fp = fopen(fn.c_str(), "r");
-	if (!fp)
+	std::ifstream fp;
```
sjanzou, line 3451: Would have to be tested for any file access issues using the PVRPM script that uses these functions

src/invoke.cpp
```diff
@@ -4007,18 +4010,19 @@ void lhs_threaded(lk::invoke_t &cxt, wxString &workdir, int &sv, int &num_sample
 	if (wxFileExists(workdir + "/LHS.ERR"))
 	{
 		err_msg = "LHS error.  There could be a problem with the input setup.";
-		FILE *ferr = fopen(wxString(workdir + "/LHS.ERR").c_str(), "r");
-		if (ferr)
+		std::ifstream ferr;
```
sjanzou, line 4013: Changes would have to be tested with LHS sampling in PVRPM code for any issue with file access

dguittet:
@sjanzou Thanks for looking through the changes. Do you or @janinefreeman have PVRPM code handy to run?

src/invoke.cpp
bje-: There really should be a testsuite!

janinefreeman: We agree! We're still working on catching up on our test coverage for our codebases- it's easier to add tests to sections we're actively working on, but we hope to be able to get more complete coverage through the upcoming years! If you happen to get motivated to add some tests, we'd be thrilled! =)

sjanzou: @bje- Could you please build SAM from this branch and run a stochastic simulation? (Right below the "Simulate" button on lower left-hand side of SAM case page).

src/widgets.cpp
bje-: If this PR can be approved and merged, I can do back and fix all remaining instances of `FILE *` uses.

src/invoke.cpp
dguittet: Just did this and the stochastic simulation worked! Can we merge this?

janinefreeman: @sjanzou thoughts?

bje-:
If there are any remaining concerns about this change, it's possible to (more or less) check it by inspection.

PR #336, opened 2020-06-19
Better parsing of lat,lon for NSRDB downloads, Fix #325

### LINKED ISSUE #325: Format of Lat/Long in Solar Resource Download
When a lat/long string is entered in a strange format on the Solar Resource page in the download box, a different lat/long may be downloaded with little feedback to the user. The correct value is shown in the Lat and Long boxes in the solar resource library, but that may be too subtle for some users (including me!) to notice that SAM didn't download what they were expecting.

The format that works correctly is:
34.818, -118.397

Two formats that do not get parsed correctly are:
34.818 -118.397
34.818N 118.397W

cpaulgilman:
Improve messaging when lat,lon is not valid decimal degrees. Add simple NSRDB download LK sample script.

To test, copy private.h from sam-private repo, or provide valid API keys in private.h

### DIFF
--- a/test_results_win64.csv
+++ b/test_results_win64.csv
@@ -87,7 +87,7 @@ Dish Stirling,All Equity Partnership Flip,207236624,18.860062,20.753241,;
 Dish Stirling,Sale Leaseback,207236624,19.091169,19.771194,; 
 Dish Stirling,LCOE Calculator,207236624,NA,NA,; 
 Dish Stirling,None,207236624,NA,NA,; 
-Generic CSP System,Commercial,574225408,3.711204,NA,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
+Generic CSP System,Commercial,573450304,3.715794,NA,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
 Generic CSP System,Single Owner,573450304,11.254256,12.133404,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
 Generic CSP System,Merchant Plant,573527808,10.609083,NA,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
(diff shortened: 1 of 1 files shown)


cpaulgilman:
I'm investigating why this pull request causes the Travis CI build to fail...

cpaulgilman:
@dguittet or @sjanzou: I think the check is failing because the compare_lk_test_output.py script is comparing results to an old version of test_results_win64.csv from when I created the solar_resource_latlon325 branch. I'm not sure how to fix that. Do you have suggestions? Don't know whether to merge patch into my branch to update the .csv file, or some other approach?

Thanks,
Paul.

dguittet:
The workflow is to update the `test_results_win64` in your branch by running the test script and commit the changed results after checking that they make sense. This helps make sure that the current default configurations are running as expected on your feature branch, and then patch after merging.

@brtietz could you take a look at the error message? https://travis-ci.org/github/NREL/SAM/builds/700224518#L6276 

Also, it looks like both SAM/patch and SAM/develop are both failing on Travis right now. Would that affect the comparisons?

brtietz:
The issue is that the expected results have "fail" instead of an annual energy in the commit being tested: https://github.com/NREL/SAM/blob/1086eddbd080773a4f4506541091e163aa8e440c/test_results_win64.csv#L124

I didn't expect us to commit things if the expected results failed, so I didn't write a very good error message. If this is common I can update it.

If that's expected (it appears to also be true on patch right now) then it shouldn't block your merge.

brtietz:
This is super weird: it looks like Travis is running a [newly created merge commit](https://github.com/NREL/SAM/commit/ccdd297309ff86e43cf9c54b24528fbf7a1715b8), such that the expected results match Steve's latest commit to patch, not the correct results listed in the changed files on the PR.

test_results_win64.csv
```diff
@@ -87,7 +87,7 @@ Dish Stirling,All Equity Partnership Flip,207236624,18.860062,20.753241,;
 Dish Stirling,Sale Leaseback,207236624,19.091169,19.771194,; 
 Dish Stirling,LCOE Calculator,207236624,NA,NA,; 
 Dish Stirling,None,207236624,NA,NA,; 
-Generic CSP System,Commercial,573450304,3.715794,NA,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
+Generic CSP System,Commercial,574225408,3.711204,NA,; Notice: time 0.00 { Generic solar model 2 }:\ The interpolation code must be 1 (interpolate) or 2 (nearest neighbor)The input value was 0, so it was reset to 1
```
brtietz, line 90: @cpaulgilman I was mistaken in the SAM team meeting yesterday, it looks like the actual results match the _old_ values for this line:

```
Large change in ANNUAL_OUTPUT for Generic CSP System Commercial. Expected 574225408 / Actual: 573450304
```

Can you confirm that you had the latest patch ssc checked out when you ran test_script_ow.lk? If so, I'm ok to merge this into patch and fix this on the patch branch.

cpaulgilman: Yes, I ran the script using the latest patch version.

Issue #343, opened 2020-07-01
Array defaults - check all 


sjanzou:
Example of array defaults that need to be checked are utility rate timestep array inputs for buy and sell rates
![image](https://user-images.githubusercontent.com/6498311/86224092-aabef400-bb45-11ea-8b95-a684f42f92ec.png)

See https://sam.nrel.gov/forum/forum-general/3030 for user reported error and current workaround.

brtietz:
Looks like this was fixed in https://github.com/NREL/SAM/pull/344. Is there anything else to do for this issue?

cpaulgilman:
We fixed the issue with not being able to enable and disable time series buy and sell rates separately, but there is a larger issue with default values of all array inputs that still needs to be fixed. All default values should be of the correct length to avoid simulation errors when a user enables an input that is an array and whose default value is the wrong length. This is an issue with inputs that are 1) arrays, matrices, or tables, 2) disabled by default, and 3) default value is the wrong length.

PR #344, opened 2020-07-04
Update utility rate flat and corresponding defaults to address SAM #343

sjanzou:
Add separate enable for time step buy rates in UI and in cmod_utilityrate5


cpaulgilman:
Thanks for this fix. I've updated Help to describe the separate enable time series buy/sell rate check boxes.

image

PR #345, opened 2020-07-10
Eliminate warnings about checking references for NULL

bje-:
The C++ standard guarantees that a reference cannot be `NULL`. It must always refer to a valid object. Perhaps what the original author wanted was a pointer to object so that it can be set to `NULL`?


bje-:
Ping?

Issue #348, opened 2020-07-16
Update wind turbine library


cpaulgilman:

mjprilliman:
Description: Update Wind Turbine Library Excel file with corrected power curves and rotor diameter values based on spreadsheet provided by PNNL. Also includes new turbine technologies not yet defined in the SAM Library

PR #349, opened 2020-07-17
Wind Turbine Library Updates (Issue #348)

mjprilliman:
Updated wind turbine library to include rotor diameter and power curve changes. To be included in Patch 3.


janinefreeman:
Closes #348

PR #374, opened 2020-08-31
ensure itc max values are 1e38 for all relevant btm configurations

brtietz:
Correct ITC max values as per our email conversation.

We might want to wait for Travis testing results to see if this changes the LCOE for those configurations, I haven't run it locally.


brtietz:
Thanks! Looks like the LCOE did change for these configurations, I'll run the test script locally and push the new results.

PR #381, opened 2020-09-17
Batt dc ac sizing

### LINKED ISSUE #375: Battery DC vs AC values for Size
For a 200 kWh / 100 kW AC-connected battery, if I size it as DC values, I get these:
 
image

If I size it as AC values, then I get these:
 
image

It looks like the first one makes sense but the second one doesn’t. If the battery is AC-connected then the AC -> DC inversion during charging should mean the max charge AC > max charge DC ~  96 kWdc.

----------------

I fixed the PySAM version of the sizing function here: https://github.com/NREL/pysam/pull/58 which describes:

1. Expected sizing function behavior:

    - AC: the max charge and discharge power on the AC side is the `desired_power` input; the max dc charge power is lower while the max dc discharge power is higher due to conversion efficiency losses
    - DC: the max charge and discharge power on the DC side is the `desired_power` input; the max ac charge power is higher while the max ac discharge power is lower due to conversion efficiency losses

    where the conversion efficiency losses are:
     - AC: `batt_ac_dc_efficiency` and `batt_dc_ac_efficiency`
     - DC: `batt_dc_dc_efficiency` * inverter efficiency

mjprilliman:
Fixed AC sizing options to accurately reflect DC and AC sizing for charge and discharge
Close #375


dguittet:
@mjprilliman This looks good so far! I think the Nominal bank capacity and Maximum charge/discharge powers in AC and DC look right. 

Looking at the GUI, there are some additional inputs there versus pysam that will require additional features. The pysam battery sizing should be updated to have the full functionality you're introducing here.

1. The code originally has `batt_C_rate_max_charge` equal to batt_C_rate_max_discharge`:

    ```
    batt_C_rate_max_discharge = batt_bank_power / batt_bank_capacity;
    batt_C_rate_max_charge = batt_C_rate_max_discharge;
    ```

    But this is only true when the `Desired bank power` size is in DC. When it's in AC, the max charge and discharge power in DC can be quite different. For example:

    image

    Here the Max C-rate of discharge of 1111/4444 = 0.25 makes sense, but Max C-rate of charge should be 900/4444

2. In the GUI (but not in the pysam code I sent you), for AC-connected batteries, there are separate AC to DC and DC to AC conversion efficiency inputs, but when `Desired bank power` size is in DC, then the AC to DC efficiency doesn't appear to be used.
    
    image

    Here the battery is AC-connected, so there is a battery inverter that converts between the battery (a DC power source) and the AC bus. The charging of the battery requires inversion from AC to DC so the Maximum charge power (DC) should be 500 instead of 1000. This will then effect `batt_C_rate_max_charge`. Maybe `batt_C_rate_max_charge` and `batt_C_rate_max_discharge` should be calculated towards the end of the function rather than near the beginning as it is now.

dguittet (approved):
Looks great! I think it's good to merge, thanks!

PR #406, opened 2020-10-21
Add transient thermal model option to UI

### LINKED ISSUE #380: Add transient thermal model option to the UI, documentation
Transient thermal model (Prilliman, 2020) included in the SSC, need to include option to choose it in SAM UI. Also update documentation with details of transient thermal model.

mjprilliman:
-Add unit mass input that is only editable when time step is less than 20 minutes
-Available in CEC (Database and User specified) and SNL module models
-Additional inputs to pvsamv1 compute module (see ssc pull request)

-Closes #380


mjprilliman:
@cpaulgilman I had to add a variable to the solar resource page to access the `step` output from the `wfreader` call. This changes a lot of default files, is that going to be an issue?

cpaulgilman:
@mjprilliman Adding the step variable should be fine. Thoughts:

* I added a time step (converted to minutes) numeric widget to the Solar Resource page under Elevation in 2ff4abe8c. That adds a `ui_step_minutes` variable for the converted value.

* Are the extra variables `cec_step`, `6par_step`, etc. variables necessary? A simpler approach might be to put `enable('cec_transient_thermal_model_unit_mass',value('step') <=1200);` in the `on_load{"CEC Performance Model with Module Database"}` function and removing the `on__change{'cec_step'}` function. My approach would not work if there were a way to change the value of step while on the Module page. Your approach might be more robust -- it's your call on how you want to do that. We can manage the additional defaults.

mjprilliman:
@cpaulgilman I'm fine with the second approach. I don't see any reason why we should allow step to be changed to a value that is different what is in the weather file right now. I will make those changes to the step check just to avoid having extra variables where they are not needed.

Issue #405, opened 2020-10-20
Scroll bars on long input pages don't always work


cpaulgilman:
For long input pages and results pages, especially ones with expanding panels, the scroll bar does not always show the bottom of the page. Workaround is to switch to a different page and back.

Some examples:

Financial Parameters page for Single Owner:

![image](https://user-images.githubusercontent.com/27711933/96650962-ae3edc80-12e8-11eb-8ba8-9d42075dfc30.png)

Battery Storage page for PV Battery:

![image](https://user-images.githubusercontent.com/27711933/96651131-01189400-12e9-11eb-9a0d-1bb345c8994e.png)

Cash Flow tab on Results page:

![image](https://user-images.githubusercontent.com/27711933/96651247-43da6c00-12e9-11eb-8526-b2683876f0d3.png)

sjanzou:
Steps to reproduce effect:

1. Create generic battery, sales leaseback configuration
![image](https://user-images.githubusercontent.com/6498311/96835170-73c86300-1400-11eb-9c77-ffd2eb2d0e87.png)

2. Open Battery Cell and System page
![image](https://user-images.githubusercontent.com/6498311/96835241-8cd11400-1400-11eb-97de-c4434d56c255.png)

3. Scroll down
![image](https://user-images.githubusercontent.com/6498311/96835291-9fe3e400-1400-11eb-9be6-4ce7e34c3c95.png)

4. Return to Power Plant page
![image](https://user-images.githubusercontent.com/6498311/96835347-b427e100-1400-11eb-9280-e4a8f5f291e2.png)

All widgets moved down on page...

sjanzou:
PR #408 fixed input page scrollbar issue but cash flow scrollbar issue remains
![image](https://user-images.githubusercontent.com/6498311/96950849-7aaab080-14a8-11eb-874e-b939f39085a0.png)

PR #408, opened 2020-10-22
Fix scrollbar issue

sjanzou:
Fixes widgets going off page bottom but resets to top of page. Cashflow also works for BOM and FOM configurations

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -1110,8 +1110,12 @@ void ResultsViewer::Setup( Simulation *sim )
 		m_depreciationTable->Thaw();
 	}
 
-	if ( m_cashflow.size() > 0 ) ShowPage( PAGE_CASH_FLOW );
-	else HidePage( PAGE_CASH_FLOW );
+	if (m_cashflow.size() > 0) {
+		ShowPage(PAGE_CASH_FLOW);
+		m_cashFlowTable->GetParent()->Layout();
+	}
+	else 
+		HidePage( PAGE_CASH_FLOW );
 
 
 	CreateAutoGraphs();


brtietz (approved):
The fix looks great! Is there a reason to leave so much commented-out code? I generally find git history more helpful since the commit message provides good context on why things were changed or removed.

mjprilliman (commented):
I'm still seeing some issues on the battery example with the bottom of the page cutting off. It seems that when you open one collapsible tab the page appropriately resizes, but upon opening a second tab the page size cuts off the bottom again and the top shifts down. 

I can note that in my testing for the ME pages programatically opening the collapsible tabs by manipulating the collapsible var used in the startup page does seem to automatically fix the page size each time. Not sure how helpful that is for the battery case but something I discovered.

cpaulgilman (approved):

src/inputpage.cpp
```diff
@@ -138,8 +138,11 @@ ActiveInputPage::~ActiveInputPage()
 
 wxRect ActiveInputPage::ScaleRect( const wxRect &r )
 {
-	return wxRect( (int)(r.x*m_scaleX), (int)(r.y*m_scaleY),
-		(int)(r.width*m_scaleX), (int)(r.height*m_scaleY));
+	// testing
+//	return wxRect((int)(r.x * m_scaleX), (int)(r.y ),
+//		(int)(r.width * m_scaleX), (int)(r.height * m_scaleY));
```
dguittet, line 143: please remove unused

src/inputpage.cpp
```diff
@@ -318,6 +321,8 @@ void ActiveInputPage::OnPaint( wxPaintEvent & )
 			if (!objs[i]->IsVisible())
 			{
 				rct = ScaleRect(objs[i]->GetGeometry());
+				// testing
+				//if (i == 0) rct.y = 0;
```
dguittet, line 325: please remove

dguittet (changes requested):
Thanks for fixing it so quickly! Agree with @brtietz that unnecessary comments should be removed. If there is extra information you'd like to track, perhaps adding them as comments to this Issue / PR will be a good reference instead.

sjanzou:
Changes requested made and scroll bar issue in #405 resolved.

sjanzou:
Address all issues and closes #405

sjanzou:
Closes #405

PR #409, opened 2020-10-22
Sam 399 fix replacement costs

brtietz:
Remove ssc variable batt_replacement_schedule. In lieu of this, use the percentages in batt_replacement_schedule_percent. Update all of the financial compute modules to account for this. Removes input from the battery cell and system page, and defaults from all relevant configurations.

Should be merged at the same time as: https://github.com/NREL/ssc/pull/464


cpaulgilman (approved):
Looks good. I added a function to versions.lk with a message to alert users of the change.

dguittet (approved):

brtietz:
The travis tests proved their worth again here: it looks like the previous code was applying the battery replacement costs a year late. Also, it looks like I stashed instead of committed some code on ssc, so there'll be another push on this one shortly with the additional fix.

dguittet:
image

PR #411, opened 2020-10-27
Update Battery defaults

### LINKED ISSUE #323: Battery defaults-- thermal, voltage & dispatch
Battery defaults for the thermal, voltage and perhaps dispatch models need to be updated and made consistent across all configurations.

1. For the thermal model, the model has been updated to resolve some of the previous model's instabilities. This allows us to now reevaluate the heat transfer coefficients, energy density, etc parameters.

2. For the voltage model, these were inaccurate for several chemistries (lead acid) and should be updated with data of representative batteries on the market.

3. For the custom dispatch option, it could be interesting to have a default pulled from the REopt sizing function. We'll have to change some settings perhaps to get a non-zero recommended battery size).

--------

Battery defaults per chemistry should be available both in LK and via PySAM's `default` function. A new doc page for `BatteryStateful` should be linked from the main module help page.

dguittet:
1. Update Lead acid, Li-ion NMC-Graphite and LFP-Graphite default values.
    - gBattery ESL 100-12 (C10) for lead acid
    - Sony Fortelion   IJ1001M for LFP
    - Kokam SLPB70460330H or similar for NMC

2. Move these values into JSON files in deploy/libraries/BatteryStateful:
    - BatteryStateful_LFPGraphite.json
    - BatteryStateful_NMCGraphite.json
    - BatteryStateful_LeadAcid.json

3. GUI loads these using a new lk function `read_json`
    - Other chemistries' default values are still hard-coded in `Battery Model.txt`

4. PySAM.BatteryStateful has these as new configurations for `default()`

Thanks @brtietz for fitting the NMC defaults from the Kokam test data. I know that process isn't 100% complete, but fairly close, so we can update in the future as well.

closes #323


sjanzou (commented):
@dguittet, does every configuration with a battery reads in the json before or after reading defaults? what is the speed impact? is there a library viewer to check settings (e.g. like the csv files for the inverters and modules)?

dguittet:
@sjanzou It's only when you use the drop down menu to change the battery chemistry at the top of the Battery Cell and System page. When you create a case the defaults are loaded from the defaults files not from this battery library.

Thanks for reminding me I haven't changed the defaults in the default folder-- which I can do now. What is the shortcut for updating defaults from the current case again?

The speed difference in the GUI is unnoticeable for me since it's such a small set of values.

There's no library viewer. Could be a nifty thing for the future if there were a full library.

cpaulgilman:
Shift-F10 updates the defaults from the current case, but *please use this carefully and as little as possible*. It updates all of the defaults for the current configuration, which makes it easy to accidentally change a default you didn't mean to. The preferred (although more tedious) way to update defaults is to use the defaults manager in the SAM development environment where you can see what variables you are changing.

For example, if you expand one of the input page panels like "Battery Lifetime" on the Battery Storage page before pressing Shift-F10, the default value of the collapsible_var `battery_lifetime_shown` will be set to 1.

brtietz (commented):
Do you have references for the new Cp and h values? h looks reasonable to me, Cp is a lot higher than what I've seen in the literature.

cpaulgilman (approved):
After I changed the folder name in /libraries, the UI works: Battery defaults load from JSON and simulation runs.

dguittet:
@cpaulgilman Thanks! I've updated the defaults. I used the output from Shift + F10 to help diff what variables I needed to change in the defaults manager. Sorry for the folder misnomer, I had forgotten to push the fix to that.

@brtietz Regarding Cp, I don't have a citation unfortunately as I was going off a recommendation by Matthew Keyser that for a full battery system, there's ~35% overhead from the cell's Cp.

dguittet:
@brtietz @sjanzou Travis tests are passing-- ready for re-review, thanks!

I'll resolve the `Battery Lifetime.txt` conflict before merging, as it's due to a simple typo fix in https://github.com/NREL/SAM/commit/487dc1fc90a30086c6bd6fac58160d88ed31b5d0

brtietz (approved):
Test results look good, nice work!

dguittet:
@brtietz I had to redo the battery lifetime matrix for NMC since the previous look up table was really under-predicting cycle loss. I based it on the analytic NREL lifetime model. Unfortunately, there isn't a perfect mapping from `min(QLi, Qpos, Qneg)` to `min(Qcal, Qcyc)`. And besides that, there are differences in what is meant by `DOD`, how the cycles are calculated, etc between the two models that make the mapping rough to start.

In any case, the code for how I generated the new lifetime matrix in the sam-analysis folder and is called analytic_lifetime_model.py. We can have a follow up discussion about this if you're interested.

For Lead acid, I changed the default to use the voltage table. The voltage model was having a difficult time fitting to the shape of the discharge curve; and once fit was resulting in really bad hysteresis on my test case and resulting in a RT efficiency of ~20%.

sjanzou (approved):
Some of these changes may need to be manually merged and then the results file will need to be checked for messages about inverter sizing, etc.

dguittet:
Results file here: https://github.com/NREL/SAM/blob/fbc0098f0521eed76792267d480bcb79c2e09185/test_results_win64.csv You can see no new error or warning messages. The size of the batteries weren't changed for the defaults, just the cell and system properties.

PR #412, opened 2020-10-27
Inverter Temperature Derate based on Pmax

mjprilliman:
-Changed inverter derate table to be based on ratio of max output power to rated power rather than efficiency
-Update Y-axis label to reflect the ratio rather than the percentage
-Update default inverter temperature derate table to reflect more realistic examples (resource provided by Will Hobbs from Southern Company: https://files.sma.de/downloads/Temp-Derating-TI-en-15.pdf)
-Related NREL/ssc#471, closes NREL/ssc#442


mjprilliman:
@cpaulgilman There will be some documentation changes based on the improvements in this PR most likely.

cpaulgilman:
Thanks for the documentation heads-up.

cpaulgilman (approved):
UI looks good. Table is a bit tricky to use, but functional enough once you understand how it works. Is there a maximum limit to the number of columns in the table?

mjprilliman:
Right now I think the labels stop after 9 columns (four different temp-slope combos). But the editor allows more columns to be generated without the appropriate columns (just called 10 and 11). Should a max number of columns be enforced and if so how?

cpaulgilman:
I think it's OK not to enforce a maximum -- we can rely on an explanation in Help. Perhaps we could add a UI label with text like "Table supports up to four temperature - slope pairs in each row." either above the table or below it somewhere next to the **Update plot** button.

Issue #393, opened 2020-10-01
SAM Error: Could not inflate stream: Unexpected EOF in underlying stream.


cpaulgilman:
Opening [this file](https://github.com/NREL/SAM/files/5315106/could-not-inflate-stream-error.zip)  in SAM 2020.2.29 for Windows results in error:

![image](https://user-images.githubusercontent.com/27711933/94869515-56543a80-03fa-11eb-803c-a929bd136766.png)

Also reported by another user here:

https://sam.nrel.gov/forum/forum-general/3146

From Brian:

> I pulled this up in the debugger and stepped through it a bit, and came up with the possibly more helpful error message:
>
>16:52:50: discrepancy reading in values from project file: 0 not found, 0 wrong type, 822 read != 824 in config
>
>Looks like maybe there are issues for cases where we adjusted the defaults between patches? Can you get any info on what case(s) are in the file so we have a better shot at reproducing it?

sjanzou:
@cpaulgilman, @brtietz, the discrepancy is from the latest version built from develop for both of the Flat Plate PV / Residential cases in the project file: (code to list missing variables in SAM_393 branch)
02:47:55: discrepancy reading in values from project file: 0 not found, 0 wrong type, 822 read != 824 in config
02:47:55: Flat Plate PV, Residential configuration variable ur_nm_credit_rollover missing from project file
02:47:55: Flat Plate PV, Residential configuration variable ur_nm_credit_month missing from project file
02:47:55: discrepancy reading in values from project file: 0 not found, 0 wrong type, 822 read != 824 in config
02:47:55: Flat Plate PV, Residential configuration variable ur_nm_credit_rollover missing from project file
02:47:55: Flat Plate PV, Residential configuration variable ur_nm_credit_month missing from project file
02:47:55: Error: Can't read inflate stream: unexpected EOF in underlying stream.

The two variable ur_nm_credit_rollover and ur_nm_credit_month will need to be added to the Version upgrade script.

The stream error is caused by a parametrics quick setup read in the second case that will, hopefully, be resolved soon.

sjanzou:
Add return true if hit end of project file when reading in parametrics – reads in second case from user – go through problematic case: 
![image](https://user-images.githubusercontent.com/6498311/97297376-918a3380-1817-11eb-9ef8-8448712f00c0.png)

sjanzou:
[...]

PR #413, opened 2020-10-29
Fix SAM issue 393

sjanzou:
Adds additional project file checks and fixes issues with duplicate parametric outputs that corrupt SAM project files


brtietz (changes requested):
@sjanzou I think there were some merge issues here. All of the batt_replacement_schedule variables and the GUI option were removed in https://github.com/NREL/SAM/pull/409. Is there a reason they needed to come back?

cpaulgilman (approved):
This looks good Steve. I pushed a commit to revise the error message:

image

sjanzou:
> @sjanzou I think there were some merge issues here. All of the batt_replacement_schedule variables and the GUI option were removed in #409. Is there a reason they needed to come back?

Not sure why these were updated... Seems like when I merged develop into #393 these got updated for some reason... I will remove from this pull request.

sjanzou:
Revert changes per @brtietz and run 419 tests and verify defaults and forms from develop and close #393

brtietz:
@sjanzou the state of this branch as of 52a789c looks good to me, but it didn't get merged into develop (closed instead of merged). Is that right?

sjanzou:
Closes #393 as before

Issue #383, opened 2020-09-21
Array DC Power and DC-connected battery


cpaulgilman:
For DC-connected batteries, the "Array DC power" output variable is the combined power of the battery and array, so it actually actually the DC power at the inverter input. It would be useful to show the array power and battery power as separate variables, or to rename the variable.

Results for DC-connected battery:

![image](https://user-images.githubusercontent.com/27711933/93817092-4aec5c80-fc0d-11ea-9f38-b28c2d3d2a0e.png)

Same system, but AC-connected battery:

![image](https://user-images.githubusercontent.com/27711933/93817142-5dff2c80-fc0d-11ea-8760-bb7f99f53e35.png)

cpaulgilman:
Also, variable names like "Electricity to grid from system" are confusing because the battery is part of the system, but the variable does not include electricity from the battery. However, there is not a good alternative to "system" that works for both PV Battery and Generic Battery configurations, and possible future implementations such as Wind Battery.

cpaulgilman:
Fixed in https://github.com/NREL/ssc/commit/ce96913e5c228bb96b702211269a8d0c8efd571d

Issue #370, opened 2020-08-17
Electricity Purchases for Front-of-Meter Battery Configurations


cpaulgilman:
For PV Battery / Single Owner and Generic Battery / Single Owner, battery charging power is treated as an operating expense even when the option on the Electricity Rates page is "No Electricity Rates". 

https://sam.nrel.gov/forum/forum-general/3108

cf_utility_bill should be an array of zeros when en_electricity_rates = 0 to help communicate that electricity purchases for battery charging are not being treated as an operating expense.

cpaulgilman:
Also, for Generic Battery / Merchant Plant configuration, there is no Electricity Rates page, so it is not possible to charge the battery at retail rates, but the cash flow table includes the variable for electricity purchases:

![image](https://user-images.githubusercontent.com/27711933/93534501-09e90500-f8fa-11ea-9d71-709a0ae9fcdd.png)

cpaulgilman:
1. For PV Battery and Generic Battery with PPA Single Owner, when en_electricity_rates = 1, the buy all / sell all metering option (ur_metering_option = 4 ) is the only valid option.

2. For the Merchant Plant and other PPA financial models (partnership flip, sale leaseback), we assume en_electricity_rates = 0, so it is not possible to enable retail rates for battery charging power. Cash flow should only dispay cf_utility_bill for Single Owner model.

cpaulgilman:
@brtietz Thanks for the fix so far. Why is annual energy changing with the electricity rates enabled/disabled option? My impression was that this option determines whether the electricity for charging the battery is treated as negative revenue or an operating expense, so would affect dollar amounts in the cash flow but not kWh. The behavior seems to be the same for manual and automated dispatch. For automated dispatch, I would expect battery dispatch to depend on the electricity rate option.

brtietz:
Charging from the grid (in either case) shows up as a negative number in gen, which is summed for annual energy. There is a function in common_financial: https://github.com/NREL/ssc/blob/c8f6e2e0f7859fc2f30a89162d02c1042ea7aec7/ssc/common_financial.cpp#L3208
which zeroes out these negative numbers by adding grid_to_batt back to gen as a preprocessing step, but only when the utility rate is enabled. The solution on the branch above skips this step for disabled rates, since en_electricity_rates is zero.

Do you know of an existing SSC var that could be used for a negative revenue for this case?

cpaulgilman:
[...]

PR #417, opened 2020-10-29
Sam 370 correct grid charge expenses

brtietz:
Pairs with SSC PR: https://github.com/NREL/ssc/pull/474

Covers both #383 and NREL/SAM#370

- Change metrics to cover name changes from SSC
- Change cashflow to include new breakdown for SingleOwner
- Run API autogen for the name changes, and pick up a lot of changes from develop as well

Plan is to have extensive review and polishing (especially of names and variable labels) prior to merge. We also need to decide if we only want to apply the PPA cashflow changes Single Owner or to all front of meter modules.


brtietz:
@cpaulgilman I included your name and ordering recommendations from the meeting yesterday, thank you! One note is that curtailment is already accounted for in "Energy to grid" and it's sign is different from "Energy from grid" so I'm torn about it's current location in the cash flow sheet. Any thoughts?

cpaulgilman:
Thanks, @brtietz. In the cash flow LK script, we can change the sign of the display value of the variable in the cash flow.

But, it seems like we should be careful about sign convention in SSC.

brtietz:
Closed with https://github.com/NREL/SAM/commit/dd650870125da5fdb8d037173422a510c42d261a

Issue #363, opened 2020-08-05
Remove Dish Stirling & Direct Steam Power tower from Startup Script


janinefreeman:
The dish stirling and direct steam technologies in SAM have become outdated and do not have current funding support to continue updating, so we are planning to remove them from the SAM user interface for our fall 2020 release.

dguittet:
@mjprilliman @sjanzou How to handle blank case with no tabs? Or an old version SAM project file that doesn't have these removed?

cpaulgilman:
@mjprilliman, What branch did you use for your work on this?

mjprilliman:
remove_csp_technologies. 

I have it checked out right now. Looking further I am seeing some issues pop up where metrics.lk, autographs.lk, and cashflow.lk are looking for the financial model information contained in the retired technologies when opening projects from an older version. So we either need to go back to figuring out how to comment out the technologies in startup without SAM crashing or I am looking at how to delete the case as part of version upgrade but so far no success.

cpaulgilman:
These possible solutions for opening an old SAM case that contains a retired technology would require writing new functions for the Version Upgrade flavor of LK:

1. Display a `msgbox()` saying that this version of SAM cannot open the file because it contains a case with a retired model and then quit SAM. (Include text in message box to explain that if the file contains other cases with non-retired models, the user can use File->Import to open those cases.) Requires a new `quit_sam()` function.

2. Remove any cases with retired models from the file and notify the user using `message()`. If the file only contains a single case with a retired model, close SAM so that the Welcome page is displayed. Requires new `delete_case()' and `close_sam()` functions.

These approaches would mean that every time we retire a case, we have to modify the version upgrade script to handle the retired models, so more effort than simply commenting out lines in the startup script. Also, for each new version of SAM, we would have to be sure to include the code that checks for retired models in the function for that version.

mjprilliman:
@cpaulgilman Take a look at my latest push on removing_csp_technologies. I couldn't figure out how to make it into an LK function yet but I have a start on option 2. My only concern would be people losing their case data if they save the upgraded project file, already there is already a two-step prompt when saving in this situation.

mjprilliman:
@sjanzou @cpaulgilman The current solution is workable but requires adjustments to the source code for future model retirements. I haven't yet figured out a way to achieve the same behavior by changing the LK scripts so any ideas would be appreciated.

mjprilliman:
[...]

PR #428, opened 2020-11-10
Remove Dish Stirling and DSPT configs from UI

cpaulgilman:


dguittet:
@sjanzou Regarding the CMakeLists.txt, when SAMAPI_EXPORT=0, then on Unix it produces ssc.so, otherwise libssc.so. This was required for the Python C Extensions build.

When you're doing the umbrella CMake build, CMake will know automatically how to find ssc.so/libssc.so and doesn't need to know the name explicitly.

When you're doing the separate builds, that's when CMake might not be able to find ssc.so if it was produced as libssc.so, or vice versa.

Also, I think we don't actually need to specify the extension `.so`, `.dylib` in the find_library function, so we could simply use 
```
find_library( SSC_LIB
			NAMES ssc libssc
```

I'm not sure what's the best way to fix this branching behavior. 
1. Keep using SAMAPI_EXPORT=0 for building SAM separately. This has the benefit of leaving the names unchanged for the sam-private scripts. User has to rebuild for ssc vs libssc.
2. Do
```
find_library( SSC_LIB
			NAMES ssc libssc
```
So Cmake will search for both. This has a potential downfall of choosing the wrong/stale version.
3. Change all our build processes to use libssc.so (sam-private)

brtietz:
I like (1) or (3). (2) sounds like a recipe for future headaches.

cpaulgilman:
Close #363

PR #429, opened 2020-11-11
update LandBOSSE to v2.3.0.3

### LINKED ISSUE #420: Update LandBosse version
LandBOSSE Python package for Wind utility-scale cost modeling needs to be updated to newest version and tested in SSC and SAM.

@parangat94 Version v. 2.3.0.3

dguittet:
Tested for open-source SAM on Mac and Windows. Will need to test on sam-private.

@sjanzou Could you please test this on Linux for the open source build? Open a Wind - Single owner case, go to the costs page, expand the LandBOSSE window, and click the Enable model checkbox. It will open a window while downloading & installing.

close #420 

merge with NREL/ssc#485

### DIFF
--- a/deploy/runtime/python/landbosse.json
+++ b/deploy/runtime/python/landbosse.json
@@ -1,5 +1,5 @@
 {
 	"min_python_version": "3.7",
 	"run_cmd": "from landbosse.landbosse_api.run import run_landbosse;sam_dict = <input>;results = run_landbosse(sam_dict);print(results)",
-	"version": "2.2.7.3"
+	"version": "2.3.0.3"
 }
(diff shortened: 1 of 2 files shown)


sjanzou:
Note - layout a little off on CentOS7
image

sjanzou:
Popup opens with no information and then closes after 1m 20s with results - so, looks like it did run but no information reported to SAM:
image

Results after blank dialog disappears:
image

sjanzou (approved):
Working with blank dialog

sjanzou:
@dguittet  Note that download and installation information appears in terminal window from which ./SAM was started:
image

dguittet:
@sjanzou Thanks for testing and noting those UI layout issues.

Any idea why the message isn't showing up on Linux?

https://github.com/NREL/SAM/blob/d4ac6827918d1284999d5dc09bc29838dd1d3857/src/invoke.cpp#L4924

PR #438, opened 2020-11-16
Fy21 upgrade function batteries

brtietz:
Add the required changes for battery inputs to versions.lk. Specifically:

- Notify users if they were using the "replace at specified schedule" option. If they had 1 replacement per year this is a notice, >1 is a warning since we've removed support for more than 1 replacement per year.
- Convert batt_cycle_cost from a single value to an array in front of meter locations.

This is a minimal set of changes, additional notices could be provided about price signals dispatch, btm cycle costs, or utility true-up months as we're discussing on email.

A test file created with 2020.2.29 is available here: 
[2020_upgrade_tests.zip](https://github.com/NREL/SAM/files/5549633/2020_upgrade_tests.zip)

### DIFF
--- a/deploy/runtime/versions.lk
+++ b/deploy/runtime/versions.lk
@@ -81,15 +81,15 @@ version_upgrade{'2020.11.12'} = define() {
     
     // Use en_batt instead of cfg.tech to capture fuel cell configurations
     if (oldvalue('en_batt') == 1) {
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
Looks good! I made a few revisions to the message text in d01ca7db8.

PR #439, opened 2020-11-16
Added retired technologies default file

mjprilliman:
Add retired technology defaults file to avoid defaults notice on upgrade script process


cpaulgilman:
This fixes the issue with the missing default file. I'm still seeing the blank project file -- is that how we're handling files that contain only retired cases for now?

cpaulgilman (approved):

mjprilliman:
@cpaulgilman For what scenarios are you seeing empty project files? Steve's fix for that was merged I believe and I haven't seen any more instances of that behavior...

cpaulgilman:
Sorry, I must not have not completely cleaned and rebuilt from this branch when I tested. It looks good: No message about defaults, and it doesn't open a file that contains only retired models. Good to go!

Issue #372, opened 2020-08-28
Net Billing results do not show $ credits


cpaulgilman:
For the Net Billing option, monthly bill is calculated correctly based on credits earned, but dollar credit amount not reported correctly. See https://sam.nrel.gov/forum/forum-general/3122.

brtietz:
It looks like it could be more of a naming issue than a code issue. Here's what I see running the SAM file from that issue with a different weather file (Blythe, CA):

![image](https://user-images.githubusercontent.com/5530592/94735164-8ed71400-0327-11eb-8976-a1d70d7ea6b7.png)

The credits are being applied immediately during the month when they are earned. "Excess generation $ credit earned" is reserved for months where the credit rolls over to the next month (Net billing with carryover to next month):

![image](https://user-images.githubusercontent.com/5530592/94735337-d5c50980-0327-11eb-8507-38c85b1a0fdc.png)

![image](https://user-images.githubusercontent.com/5530592/94735371-e5dce900-0327-11eb-801c-1086db6b173a.png)

Thoughts on name changes that might make this more clear? Or am I missing something about the bug?

cpaulgilman:
I agree there is a labeling problem, and also think it would be helpful to modify some of the monthly output variables to help people understand how SAM calculates the monthly bill for he different billing options:

**Excess generation** (kWh) for both net energy metering options: 
 
    * sum(generation[h]) - sum(load[h]) for net energy metering options

    * sum(generation[h] - load[h]) for net billing options

    * set to zero for buy all / sell all

**Net metering credit** (kWh), for *net metering* only, kWh credit that applies to current month (includes this month's excess generation and any credits rolled over from previous months).

**Net metering credit** ($), for *net metering with $ credits* only, shows $ credit that applies in current month. 

**Net billing credit** ($), for two net billing options, shows $ credit that applies in current month, regardless of when it was earned. Or, use that for *net billing* only, with a separate **Net billing carryover credit** ($) for *net billing with carryover to next month*.

brtietz:
Finally had some time to dig into this. I like these categories, are these generally additions or replacements? Some notes are below:

With net metering **Excess generation** only fits that calculation when there is one TOU period. For example, for the SAM default rates, if sum(generation[period_3]) > sum(load[period_3]) but sum(generation[period_1]) < sum(load[period_1]) at the end of the year, the customer would still receive true-up credits for the excess generation in period 3. I would favor reporting those credits in the **Excess generation (kWh)** column.

[...]

PR #443, opened 2020-11-17
UI tweaks for utility rates

cpaulgilman:
* Revise excess generation labels per discussion in #372

* Edit monthly energy and load autograph to not show excess generation so it works for all metering option. Also add it to Host / Developer model.


brtietz (approved):
The autograph looks much cleaner! Labels look good too!

Issue #385, opened 2020-09-24
Fix hard crashes Report Generator


dguittet:
@mjprilliman

mjprilliman:
Haven't spent too much time on this but the main thing is to take some function out of the function reference as trying to call them causes SAM to crash. I'll try to keep a running list of functions here:

sector()

sjanzou:
@mjprilliman, are the steps to reproduce:
1. Open report template
2. Open object editor
3. Add sector call
CRASH

mjprilliman:
@sjanzou Yes that about covers it. Everything seems to be fine until you close the sector call (closing parentheses) and then some process tries to run it. At that point SAM freezes then hard crashes.

sjanzou:
See Pull Request #442 and attached document for fixes. 
Note that the report editor supports only the pdf device context and can draw on, place images, and render tables. There is currently no direct support for the plPlotCtrl on any widget on the report editor and the crash was caused by non-implemented functions being called in the script window or the report editor.
[SAM_385.docx](https://github.com/NREL/SAM/files/5559734/SAM_385.docx)

PR #444, opened 2020-11-18
SAM issue #385

sjanzou:
Remove unsupported report functions causing hard crash in report script window


cpaulgilman (approved):
Shouldn't the unsupported functions not be available or listed in Help? 

The current solution is fine for internal use.

mjprilliman (approved):

sjanzou:
> Shouldn't the unsupported functions not be available or listed in Help?
> 
> The current solution is fine for internal use.

@cpaulgilman, sorry for the confusion, the unsupported functions are removed from the Help in the report editor window, too!

PR #445, opened 2020-11-18
Add ReOpt Error messages

dguittet:
- Add message to catch timesteps that aren't 15, 30, or 60 minutes
 - Add message to catch mismatched timesteps between weather file and load input
 - Use ReOpt's "time_steps_per_hour" to do dispatch on subhourly time scales rather than repeating hourly value per hour.

Pairs with https://github.com/NREL/ssc/pull/490

### DIFF
--- a/deploy/runtime/ui/Battery Model Simple with REopt.txt
+++ b/deploy/runtime/ui/Battery Model Simple with REopt.txt
@@ -682,7 +682,7 @@ equations{'batt_simple_meter_position'} = define()
 	return batt_meter_position;
 };
 
-2723
+3605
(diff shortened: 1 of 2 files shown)


mjprilliman (approved):
The dispatch settings are conserved for an early ReOpt exit while also allowing ReOpt to reset the array size based on the length of the weather file and load array

mjprilliman:
@dguittet I can't list you as a reviewer as it's your PR but if you could take a final look that would be great.

dguittet:
@mjprilliman Looks great! Thanks for the help

Issue #442, opened 2020-11-17
Results page data tables for demand charges do not display correctly


cpaulgilman:
## Summary of two issues:

* For residential configs, "Electricity Demand Data by Period" section does not appear in Data Tables when demand charges are enabled on Electricity Rates page, but it does when demand charges are disabled.

* For all configs with electricity rates, need two sections on the Data Tables tab to show 1) fixed monthly demand charges and 2) TOU demand charges.

SAM file from Beta 2020-11-17:

[demand-charge-results-tables.zip](https://github.com/NREL/SAM/files/5555660/demand-charge-results-tables.zip)

## Detail

For the "residential" case in the attached SAM file, there are fixed monthly and tiered/TOU demand rates specified on the Electricity Rates page:

![image](https://user-images.githubusercontent.com/27711933/99434133-4b2e6e80-28c3-11eb-96a2-1a305f4abe7d.png)

But the Data Tables tab does not have a section for demand rates:

![image](https://user-images.githubusercontent.com/27711933/99434258-6c8f5a80-28c3-11eb-91f0-302c4bb7c82e.png)

When demand charges are disabled for the residential case, then there is a demand rates section:

![image](https://user-images.githubusercontent.com/27711933/99434456-ac564200-28c3-11eb-8c54-01a0ecda7ece.png)

For both residential and commercial cases, the "Electricity Demand Data by Period" table only shows demand charges from the TOU/Tiers demand charge rate table, but not for the fixed monthly demand charge table. It also does not show tiers, only TOU periods. There should be two separate Data Tables sections: One for "Demand Charge by Month" and one for "Demand Charge by Tier and Period".

![image](https://user-images.githubusercontent.com/27711933/99435685-52ef1280-28c5-11eb-86cb-378fdcd45bad.png)

brtietz:
@cpaulgilman I'll leave the first issue to @sjanzou, but I noticed for the second issue that the monthly demand charges show up in the monthly data:

![image](https://user-images.githubusercontent.com/5530592/99462575-3799fc80-28f1-11eb-96e2-c27dc1cd504d.png)

Is that good enough or should they be moved into their own group?

sjanzou:
Note - same issue in 2020.2.29 release when enable demand unchecked:
![image](https://user-images.githubusercontent.com/6498311/99528642-d4917f80-295b-11eb-9d1f-9ade78eb041a.png)

Still working through the issue in the SAM_442 branches of ssc and SAM...

[...]

PR #447, opened 2020-11-19
Fix results browser ignoring matrices with one column partial fix to #442

sjanzou:


brtietz (approved):

cpaulgilman (approved):
Thank you for getting that working.

For a future release, we may want to report demand charges in in separate tables to show 1) tiered monthly and 2) tiered TOU charges.

PR #421, opened 2020-11-06
Sam 392 pv battery cost updates

brtietz:
Cost updates for PV and batteries based on:

https://github.com/NREL/SAM-documentation/blob/master/Defaults%20References%20and%20Tools/pv_and_battery_source_sheets/Inputs%20for%20SAM.xlsx for PV and
https://github.com/NREL/SAM-documentation/blob/master/Defaults%20References%20and%20Tools/pv_and_battery_source_sheets/BESS%20and%20PV%2BBESS%20Costs%20for%20SAM%202020-11-02.xlsx

Increase default PV sizes to 50 MW for utility scale and 500 kW for commercial. Re-order battery cost inputs in order to make it clear that $/kW already includes BOS and installation for the battery.

PR #450, opened 2020-11-20
Reduce ITC and PTC to 2021 values

brtietz:
Most technologies got a reduction from 26% to 22%. It looks like we missed a few last year (biopower, geothermal, some CSP) so these have been reduced from 30% to 22%. Source for this: https://programs.dsireusa.org/system/program/detail/1235

Additionally, reduce the wind PTC from $0.022/kWh to $0.015. Source: https://windexchange.energy.gov/projects/tax-credits This a 2020 value and might require an additional update when the 2021 values are more clear.

Update the LCOE values in test_results_win64 to reflect this. Will do Linux once these results are approved.


deploy/runtime/defaults/Biopower_Single Owner.txt
```diff
@@ -2523,7 +2523,7 @@ const_per_interest_total
 1
 1
 1
-1.82285e+06
+1822850
```
mjprilliman, line 2526: Is this format change in the default files normal? There are a lot of instances of this that must have just happened automatically

deploy/runtime/defaults/Generic Battery_Commercial.txt
```diff
@@ -1794,7 +1794,7 @@ itc_fed_percent
 1
 1
 1
-26
+0
```
mjprilliman, line 1797: Remind me why this one's going to 0?

deploy/runtime/defaults/Generic System_Commercial.txt
```diff
@@ -823,7 +823,7 @@ itc_fed_percent
 1
 1
 1
-26
+0
```
mjprilliman, line 826: Assuming zero because the technology is not defined in Generic?

mjprilliman (approved):
I'd just double check with @cpaulgilman or @tyneises on those defaults changing from scientific to integer notation as I feel like that is something that has come up in the past? I could also just be mixing that up with something different. The actual rate changes all look good to me.

deploy/runtime/defaults/Generic Battery_Commercial.txt
brtietz: Generic system's defaults assume a thermal system, which is not subject to the ITC. Some users use generic battery for a stand-alone battery, but stand alone batteries also aren't subject to it. I figure it's better to make someone who is importing an existing renewable generation profile (like I did on the webinar) increase it, than to have it be too high for the two most common cases.

deploy/runtime/defaults/Generic System_Commercial.txt
brtietz: Exactly. In fact the "heat rate" default applies the generic system might be a thermal generator.

deploy/runtime/defaults/Biopower_Single Owner.txt
brtietz: Yeah, that also happened to me on https://github.com/NREL/SAM/pull/421/files and no one thought it was a big deal. I don't know the history, so I'll leave this open until ~4 MT today in case Paul or Ty want to weigh in.

PR #454, opened 2020-11-23
Update module and inverter libraries

mjprilliman:
-Number formatting on CEC Module library update is changing I0_ref value to 0.0 (typically in range of 1e-10). 
-Not sure if it is an LK issue or in the compute module
-This is messing up the appearance of the module IV curves in the SAM interface. 

-Best fix: fix the number formatting 
-Temporary fix: set the I0_ref to an arbitrary small non-zero value


cpaulgilman (approved):
Thanks @mjprilliman for this workaround.

brtietz (approved):

PR #460, opened 2020-11-25
Fixes help page returned on Linux fixes #456

### LINKED ISSUE #456: SAM Release Candidate on Linux always goes to /index instead of the proper help page
When pressing the help button on Linux, the url sent to the default browser always goes to /index.html, rather than the help page associated with the page where the button is pressed. Paul said the correct behavior is closer to what happens on Windows where it goes to the specific page, just using the browser.

This occurs both on the open source version, and the 2020.11.23 release candidate.

sjanzou:


sjanzou (commented):
Tested on CentOS7

PR #465, opened 2020-11-25
Move MEwave macros into ME wave folder

brtietz:
The Marine Energy - Wave macros were appearing for every technology. Move them to a folder in the macros folder so they only appear for MEwave.


mjprilliman (approved):
Can you also make a folder named MEtidal with the same macros in case someone wants to use those macros for that as well? I'll double check but I've used the macros on both techs in the past.

mjprilliman:
Nice catch, I'm sure I meant to do that at some point but I must've forgotten about it.

brtietz:
I tested that just now and get this for ME Wave Report:

image

The ME Wave Compare cases gives the error:

```
Macro did not finish.
[679] referencing unassigned variable:baselineCase
```

So it looks like some modifications might be necessary to make them work with tidal.

mjprilliman:
You are right ME Wave Report does not work I must've misremembered. 

On the ME Wave Compare Cases I am getting wave and tidal cases to work so long as I click simulate on each case at least once before running the macro. 

I'm fine with just having the MEwave folder for now as that is the intended use for the macros. If there's any desire for further use it can be addressed in the patch. I'd say go ahead and merge.

PR #466, opened 2020-11-25
Library macro fixes

### LINKED ISSUE #464: System Sizing Macro Fails
For the default singleowner case, the system sizing macro fails with the following error message:

```
Getting input parameters.
Running simulation to get operating voltages.
Setting up scenarios.
Generating design suggestions  based on 25239 modules in library.
Macro did not finish.
[592] runtime exception at line 592: access violation: expected array for indexing, but found null
```

Not sure if this is due to the default size changes, or the updated module/inverter library changes.

mjprilliman:
-Add Manufacturer column back to module library
-Should resolve error with System Sizing macro
-Closes #464


brtietz (approved):
Looks great, thanks for debugging this so quickly!

cpaulgilman (approved):

PR #473, opened 2020-12-08
move puerto rico download macro to the sample lk scripts folder

brtietz:
- Move the macro from deploy/runtime/macros to samples/lk scripts for sam
- Remove all references to "macro" and the macro inputs comments
- Add hardcoded default inputs to match macro defaults
- Move URLs from webapis.conf to the lk script, since importing the url from webapis.conf automatically adds the API key, which doesn't work with the open source version

Note: these URLs can be removed from SAM-Private's webapis.conf file after this PR is merged.


cpaulgilman (approved):

brtietz:
Merged in 8e3245d31171482835ab9bd2c91fc1b16f1f82e0

PR #474, opened 2020-12-08
Add UI equation to calculate tower height from h_tower, fix #471

### LINKED ISSUE #471: Tower height does not update on Power Tower costs page
Equation missing to assign value of `h_tower` to `ui_tower_height` on Tower SolarPilot Capital Costs UI form.

> ... The tower height value in SAM is not updating. Start with the default values in SAM (create a new CSP MS power tower model) and then use the correlation in the SAM help:
>
> Total Tower Cost = Fixed Tower Costs x e ^ ( Tower Cost Scaling Exponent x ( Tower Height - Receiver Height ÷ 2 + Heliostat Height ÷ 2 ) )
>
> The SAM result matches the correlation output. Change just the receiver height and the SAM output agrees with the correlation. Same is true for heliostat height. Update the tower height and they disagree.

cpaulgilman:
* Adjust group box widths to match O&M Cost form.
* Fix alignment issues

### DIFF
--- a/deploy/runtime/ui/Molten Salt Tower Receiver.txt
+++ b/deploy/runtime/ui/Molten Salt Tower Receiver.txt
@@ -357,7 +357,7 @@ Width
 408
 Height
 3
-66
+57
 Tool Tip
 5
 0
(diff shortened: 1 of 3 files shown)


brtietz:
Should the target branch on this be patch?

cpaulgilman:
Shoot! I meant to branch this off of patch -- thanks for catching that, @brtietz.

cpaulgilman:
Replace with branch off patch PR #475

PR #475, opened 2020-12-08
Fix tower height cost issue

cpaulgilman:
* Add h_tower to UI equation for ui_tower_height input variable.
* Adjust group box widths to match O&M costs form
* Fix alignment issues

### DIFF
--- a/deploy/runtime/ui/Tower SolarPilot Capital Costs.txt
+++ b/deploy/runtime/ui/Tower SolarPilot Capital Costs.txt
@@ -2076,7 +2076,7 @@ Name
 csp.pt.cost.storage_mwht
 X
 3
-123
+141
 Y
 3
 327


brtietz (approved):

Matthew-Boyd:
Can you explain this issue to me? I see that changing the tower height on the Tower and Receiver page does not update the same-named calculated variable on the System Costs page, but do we know why this happened? The input on the Tower and Receiver page is named 'THT' and 'h_tower' is just calculated by returning 'THT'. Now we're simplifying this equation by just returning 'h_tower' instead of either 'h_tower' or 'THT', but it doesn't appear it would affect the outcome (although it does). What am I missing?

cpaulgilman:
The change that fixes this issue is on the Tower SolarPilot Capital Costs form not on the Tower Capital Costs form. The `ui_tower_height` input was not assigned on Tower SolarPilot Capital Costs,

The Tower Capital Costs form is used by the ISCC and DSPT models, which are both disabled in SAM 2020.11.29. I made the change there in case we bring those models back to life in the future. The `h_tower` variable is assigned on Tower and Receiver input page for all of the tower technologies, so there is no need for the IF statement on the costs page to determine whether to use `h_tower` or `THT`. UI forms that assign `h_tower` are: Direct Steam Tower Receiver, ISCC Molten Salt Tower Receiver, Molten Salt Tower Receiver, and MSPT receiver.

Matthew-Boyd:
Thanks, that should have been obvious. I added a commit to align one of the adjacent input boxes.

Matthew-Boyd (approved):

Issue #479, opened 2020-12-11
REopt integration not working for Generic-Battery configurations


cpaulgilman:
To reproduce:

1. Create Generic-Battery, Residential case.

2. On Battery Cell and System page, click **Get size and dispatch**. 

UI callback error:

```
Could not evaluate callback function:call_reopt->on_change
[5]: error: _s
[5]: error: _s
[4]: eval error in statement list
```

The error is caused by this line in the on_change{'call_reopt'} function in the Battery REopt Optimization callback script:

```
	wf_len = 60 / ${ui_step_minutes} * 8760;
```

For the PV-Battery configuration `ui_step_minutes` is set by an equation on the Solar Resource Data UI form, but the Generic-Battery configuration does not use that form.

One solution is to set `ui_step_minutes` in an equation on the Generic System Plant page because the options on that page determine the simulation time step.

A better solution would be to detect the simulation time step without referring to inputs on other forms to minimize risk of breaking the REopt call. As we make changes to the Solar Resource or Generic System Plant page, we may not be aware of the connection to the REopt call.

https://sam.nrel.gov/forum/forum-general/3098

cpaulgilman:
Follow-up testing in SAM 2018.11.11 r3 results in a different call-back error:

```
Could not evaluate callback function:call_reopt->on_change
[28]: error in call to 'reopt_size_battery()': Must be run from Photovoltaic case with Residential, Commercial, Third Party or Host Developer model.[13]: eval error in statement list
```

I think this is because the reopt_size_battery() LK function in the callback script requires that a latitude and longitude value be set so it can download a weather file, but for the Generic - Battery configurations, there are no latitude or longitude inputs.

A solution to would be to add lat and lon as input parameters to reopt_size_battery() so they could be set in the callback, perhaps read from UI inputs.

Another solution is to only enable REopt integration for PVWatts - Battery and PV - Battery configurations.

dguittet:
Thanks for posting and looking through why it may not be working! Unfortunately, the ReOpt sizing fxn is not meant to be used with the Generic Battery at all. It's tightly coupled with PV simulations, and will expect PV-related inputs. So I think the best solution is to remove that from the Generic Battery configurations

PR #483, opened 2020-12-15
Sam #479 remove reopt from generic batt configs

cpaulgilman:


cpaulgilman:
This should be merged into Patch, not Develop.

PR #484, opened 2020-12-15
Remove REopt from Generic Battery configs

cpaulgilman:
This fixes SAM #479

* Startup script now has functions btm storage with and without reopt.
* Remove `value_of_lost_load` from defaults for generic battery btm configs.

Also, add explanation to Help: REopt optimization is only available for PV storage BTM configurations.


dguittet (approved):
Looks good, thanks!

PR #489, opened 2020-12-30
Remove unused variables from ME O&M Costs page with invalid data types

mjprilliman:
-Invalid data types were found to not be used in the UI
-defaults for wave and tidal updated
-Also fix label hiding bug on OM cost page


sjanzou (commented):
Everything looked good - thanks for removing

PR #498, opened 2021-01-14
Added check for equal LCOE values on ME compare cases macro

### LINKED ISSUE #497: ME Compare Cases Macro fails for two cases with identical LCOE
**Describe the bug**
ME Compare Cases macro fails when comparing two cases with identical LCOE values. 

**To Reproduce**
1. Create SAM with two default ME Wave cases (make sure cases are identical)
2. Run each case to get an LCOE value
3. Navigate to macros tab, select ME Compare Cases and Run
4. See null error

**Expected behavior**
Macro should be able to prepare html for cases with identical LCOE values. 

 - OS: all (seen on Windows)
 - SAM 2020.11.29

**Additional context**
Just needs a '>=' check in the macro LK script.

mjprilliman:
-Add = check on ME compare cases macro to avoid failure when comparing two cases of identical LCOE. 
-Closes #497

### DIFF
--- a/deploy/runtime/macros/MEwave/ME Wave Compare Cases.lk
+++ b/deploy/runtime/macros/MEwave/ME Wave Compare Cases.lk
@@ -667,15 +667,15 @@ if (num_cases ==2)
 
 	if (lcoe[1]>lcoe[0])
 	{
-	baselineCase = 1;
-	advancedCase = 0;
+		baselineCase = 1;
+		advancedCase = 0;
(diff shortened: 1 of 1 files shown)


mjprilliman:
@cpaulgilman I'm going to merge this if there are no objections. This bug would probably never occur in normal use but I stumbled upon it so I just added a small check.

cpaulgilman (approved):

Issue #493, opened 2021-01-05
Revert ITC to 26%


brtietz:
Congress extended the ITC at 26% in the omnibus for two more years: https://www.solarpowerworldonline.com/2020/12/solar-investment-tax-credit-extended-at-26-for-two-additional-years/

Ensure all ITC values are 26% in the patch, vs the 22% we shipped with 2020.11.29

PR #500, opened 2021-01-15
Revert the ITC to 26%

brtietz:
Change the default federal ITC value from the expected 2021 value of 22% to the actual value of 26%: https://www.solarpowerworldonline.com/2020/12/solar-investment-tax-credit-extended-at-26-for-two-additional-years/

Closes https://github.com/NREL/SAM/issues/493


cpaulgilman (approved):
Looks good. One quibble is that I think the geothermal ITC is 10%, see https://programs.dsireusa.org/system/program/detail/658. But, we've always used the solar ITC value as default in prior versions of SAM, so I don't think that needs to change for this pull request.

PR #502, opened 2021-01-16
Batt cutoff voltage

mjprilliman:
-Add battery cutoff voltage to UI for input to voltage model
-Currently only works for dynamic voltage model (lithium ion)
-


brtietz:
First impressions on pulling up the GUI:

When I first open the "Battery Voltage" pane, the first entry for Cell cutoff voltage looks like an input:

image

but then as soon as I mouse over it, it changes to calculated:

image

We can talk more about Qfull after cutoff implementation and the calculated cell cutoff voltage field at the meeting. 

I noticed you updated the defaults, which is good, but battery parameters that vary with chemistry are also in JSON files that are loaded when the chemistry changes. Should cutoff voltage be added to these files?

Should the cutoff voltage be greyed out when the voltage table is used? It might make sense to move these to "electrochemical properties" rather than "voltage properties"

mjprilliman:
> First impressions on pulling up the GUI:
> 
> When I first open the "Battery Voltage" pane, the first entry for Cell cutoff voltage looks like an input:
> 
> image
> 
> but then as soon as I mouse over it, it changes to calculated:
> 
> image
> 
> We can talk more about Qfull after cutoff implementation and the calculated cell cutoff voltage field at the meeting.
> 
> I noticed you updated the defaults, which is good, but battery parameters that vary with chemistry are also in JSON files that are loaded when the chemistry changes. Should cutoff voltage be added to these files?
> 
> Should the cutoff voltage be greyed out when the voltage table is used? It might make sense to move these to "electrochemical properties" rather than "voltage properties"

I noticed the first issue yesterday it's an issue with the initialization in my callback script that I still need to fix. And how the value is defined is up for discussion as well. Right now, unchecked should display a calculated value of 0.66*Vnom, while checked should change to an input where the user inputs the cutoff voltage. 

I wasn't aware of those JSON files for the battery parameters so I'd refer to your judgement on that. The behavior for the voltage table was one of my discussion points for the meeting.

deploy/runtime/ui/Battery Voltage.txt
```diff
@@ -1361,7 +1756,7 @@ function cell_voltage_vs_dod()
 			count++;
 			t = i*0.1;
 			q0 = Qfull - I*t;
-			V = E0 - R*I - K*(Qfull/(Qfull-I*t)) + A*exp(-B*I*t);
+			V = E0 - R*I - K*(Qfull/(Qfull -I*t)) + A*exp(-B*I*t);
```
brtietz, line 1759: Should this be changed to batt_Qfull_cutoff? It would be great if we could get the blue line to hit 100% on the chart (assuming that matches the behavior of the battery in SSC.)

deploy/runtime/ui/Battery Voltage.txt
```diff
@@ -1359,16 +1447,25 @@ function cell_voltage_vs_dod()
...
-
+		//I = Qfull_cutoff * C_rate;
+		//I_v = Qfull_cutoff * C_rate;
 		for (i = 1; i <= stop; i++){
 			count++;
-			t = i*0.1;
+			t = i*0.10;
 			q0 = Qfull - I*t;
-			V = E0 - R*I - K*(Qfull/(Qfull-I*t)) + A*exp(-B*I*t);
+			V = E0 - R*I - K*(Qfull_cutoff/(Qfull_cutoff -I*t))+ A*exp(-B*I*t);
```
brtietz, line 1456: The formula in SSC looks like it has one more instance of Qfull than this one. Is that true or am I missing something?

PR #503, opened 2021-01-21
ReOpt VOLL & original size restoration

### LINKED ISSUE #482: Remove ReOpt Value of Lost Load Input
The VOLL is not used in battery sizing for NPV optimization, and is not used in SAM's resiliency metrics. The value doesn't do anything and should be removed.

### LINKED ISSUE #495: PVWatts-battery ReOpt sets battery capacity inputs to zero before dispaying user prompt
**Describe the bug**
Battery capacity and power inputs are set to zero even if user chooses not to override inputs:

image

**To Reproduce**
1. Create a PVWatts-battery / Commercial case.
2. On System Design page, click *Get size and dispatch*.

**Expected behavior**
Battery capacity inputs should only be overridden when user clicks Yes in popup.

2020.11.29

**Additional context**
Need to also test this for Detailed PV-battery.

dguittet:
Fix #482 where the value of lost load is not used and is confusing

Fix #495 where the original battery size is restored if ReOpt's recommendation isn't used

### DIFF
--- a/deploy/runtime/ui/Battery Model Simple with REopt.txt
+++ b/deploy/runtime/ui/Battery Model Simple with REopt.txt
@@ -1,7 +1,7 @@
 Battery Model Simple with REopt
-611
-407
-12
+589
+390
+13
 GroupBox
 
 1
@@ -18,7 +18,7 @@ Y
 0
 Width
 3
-584
+572
 Height
 3
 387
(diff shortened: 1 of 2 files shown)


cpaulgilman (approved):
Looks good. I edited the UI text and messages for clarity:

image

PR #507, opened 2021-01-22
Ensured voltage model inputs are disabled whenever iron flow battery is selected

### LINKED ISSUE #505: Iron flow battery voltage model becoming erroneously enabled
**Describe the bug**
After selecting the iron flow battery chemistry, The voltage model choice is set to the table and the voltage model inputs are disabled. Clicking the electrochemical model option does not change the value of the voltage choice variable but it does re-enable the voltage model inputs which cannot be used for the iron flow battery. They should be greyed out as long as the iron flow battery is selected. The inputs currently are greyed out again when reloading the Battery Voltage Tab. 

Expected Behavior: 
The voltage model inputs should be greyed out as long as the iron flow battery is selected. Perhaps the voltage model choice RadioChoice should be disabled as well for this battery as the electrochemical model is not an option for this battery chemistry. 

Screenshot
image

mjprilliman:
-Closes #505

### DIFF
--- a/deploy/runtime/ui/Battery Voltage.txt
+++ b/deploy/runtime/ui/Battery Voltage.txt
@@ -1218,7 +1218,7 @@ Depth of discharge [%]| Cell voltage [V]
 DataMatrix
 
 0
-4869
+4901
 on_load{'Battery Voltage'} = define()
 {
 	set_voltage_choice(false);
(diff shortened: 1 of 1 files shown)


brtietz (approved):
@mjprilliman - GUI changes look good. EDIT: Paul ignore this, the help is accurate.

PR #509, opened 2021-01-26
Add right number of commas in header of Sandia CSV

jmarrec:
Reading it up via pd.read_csv for eg will fail otherwise (without workarounds), and github can't display it because of that.

Right now:

image

Corrected:

image


cpaulgilman (approved):
I verified that SAM can read the file with this modification.

jmarrec:
Thanks!

Issue #355, opened 2020-07-24
Separate O&M Costs for PV + Battery Systems


brtietz:
See https://sam.nrel.gov/forum/forum-general/3065 about a request for O&M costs to scale with battery size, separate from PV.

From duplicate issue #169:

> Add $/kW for battery replacements to be consistent. Currently, battery replacements are specified on a $/kWh basis, while battery costs on the System Costs page are $/kW and/or $/kWh.

cpaulgilman:
See also https://github.com/NREL/SAM/issues/169

brtietz:
See also https://github.com/NREL/ssc/issues/257

brtietz:
This requires additional variables in SSC such that these O&M costs can be broken out as line items in the cash flow sheets.

sjanzou:
@cpaulgilman have you updated the UI for this layout and, if so, on what branch? Thanks!

Latest, in develop: 
![image](https://user-images.githubusercontent.com/6498311/96080500-6f3a0280-0e74-11eb-8b22-c45f681b6aa0.png)

cpaulgilman:
@sjanzou @brtietz Does this require adding a new SSC variable for battery replacement cost in $/kW (maybe use single escalation rate for both battery replacement costs) before I add an input to the UI?

sjanzou:
@cpaulgilman, yes, a new variable that will be passed to ssc from the O and M UI and used to calculate another cashflow line item in all financial models that support battery configurations. So, choose the name you like - will be used for SAM, ssc, and the Excel workbooks with equations!

cpaulgilman:
Current implementation of O&M costs is hacky (see especially fuel cell O&M costs), so we want to do a systematic refactoring of the O&M cost code before adding the new battery O&M cost as part of levelized cost of storage (LCOS) project. Notes from 10/22/2020 meeting:

* Use single $/kW value for battery O&M cost.

* Consider replacing SSC variable names like `om_capacity1` with more meaningful names like `om_capacity_battery` to make it easier for SDK users to understand variable names.

* Consolidate variable assignments into one place (common.cpp?) rather than in each financial model (singleowner.cpp, etc.) so that these variables are available to all technology and financial models as needed.

* Treat "normal" O&M costs, fuel cell O&M costs, biopower feedstock costs, and CSP fuel costs in a consistent way.

* Make sure O&M costs are shown correctly in the cash flow for all financial models, and in send-to-Excel with Equations spreadsheets.

[...]

PR #521, opened 2021-02-02
Fix capacity nameplate for Battery OM costs in Fuel Cell

### LINKED ISSUE #520: Fuel Cell battery OM costs using wrong capacity value
**Describe the bug**
The nameplate capacity used for the battery OM Costs in the PV-Fuel Cell-Battery config appears to be using the kWh capacity rather than the kW capacity. This is not the correct usage as used in the financial models. 

Expected behavior: The nameplate variable om_capacity1_nameplate should be calculated from batt_power_discharge_max_kwdc or batt_power_discharge_max_kwac as is done in the capital costs (battery_power). 

image

image

image

**Additional context**

mjprilliman:
-Changes capacity nameplate definition from kWh to kW for battery OM cost calculations
-Closes #520

### DIFF
--- a/deploy/runtime/ui/Fuel Cell O and M Costs.txt
+++ b/deploy/runtime/ui/Fuel Cell O and M Costs.txt
@@ -2447,7 +2447,7 @@ Operation and Maintenance Costs
 0
 Default
 
-916
+1142
 
 equations{ 'om_fuel_cost'} = define(){
 	fuel_price = ${fc_fuel_cost};
@@ -2469,7 +2469,16 @@ equations{ 'add_om_num_types'} = define(){
 	return 2;
 };
 equations{ 'om_capacity1_nameplate'} = define(){
-	return ${batt_computed_bank_capacity};
+	if( technology() == 'PV Battery' || technology() == 'Fuel Cell')
+			return ?${en_batt}
+				[0,	${batt_power_discharge_max_kwac}];
+	else if ((technology() == 'PVWatts Battery' ))
+		{
+			return ?${batt_simple_enable}
+				[0,	${batt_simple_kw}];
+		}
+	else
+		return 0;
 };
 equations{ 'om_capacity2_nameplate'} = define(){
 	return ${fuelcell_power_nameplate};


cpaulgilman (changes requested):
Thanks for finding this issue. The "Fuel Cell O and M Costs" UI form is only used for Fuel Cell configurations, so I think the checks for PV Battery and PVWatts Battery configurations can be omitted.

The O&M cost inputs for fuel cell are still a bit confusing because the replacement cost in $/kWh appears under the kW capacity column, but we can fix that later when we clean up the O&M variable names for https://github.com/NREL/SAM/issues/355.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
+---
```
sjanzou, line 1: Why are these files here?

deploy/runtime/ui/Fuel Cell O and M Costs.txt
```diff
@@ -2469,7 +2469,16 @@ equations{ 'add_om_num_types'} = define(){
 	return 2;
 };
 equations{ 'om_capacity1_nameplate'} = define(){
-	return ${batt_computed_bank_capacity};
+	if( technology() == 'PV Battery' || technology() == 'Fuel Cell')
+			return ?${en_batt}
```
sjanzou, line 2473: what about generic battery?

sjanzou (changes requested):
Why are the but_report and request_feature files to be merged into patch?
Why are the epw and csv files to be merged into patch - are they part of new defaults?
Was the Sandia module library updated in this pull request?

.github/ISSUE_TEMPLATE/bug_report.md
mjprilliman: I accidentally branched off of develop originally then tried to change the branch to patch here. I may need to redo this. I'm not sure why bug_report and request_feature are here.

deploy/runtime/ui/Fuel Cell O and M Costs.txt
cpaulgilman: @sjanzou This is on the Fuel Cell O and M Costs form that is only used for the Fuel Cell configurations, so no need for the configuration checks for other battery techs here.

PR #526, opened 2021-02-04
Fix #518 version upgrade for weather files and batteries back to SAM 2017.9.5

### LINKED ISSUE #518: Hard crash during weather file download in upgrade script
**Describe the bug**
When opening older SAM files, a dialog box opens to allow downloading the older weather file if it is not available. When attempting to use this dialog, SAM crashes.

**To Reproduce**
1. Open https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/SAM%202017.1.17%20with%20TOU%20rates/APS%20TOU_battery_NREL.sam with SAM 2020.11.29
2. Choose yes on this dialog:
image
(you may need to delete that file from your downloaded weather files)
3. SAM does a hard crash

**Expected behavior**
SAM shouldn't crash.

 - OS: Windows
 - Version 2020.11.29 R0

**Additional context**
It may be that the issue is I'm trying to jump too many versions. Additional testing may be needed to narrow down the issue. I haven't taken the time to gather messages through SAM debug yet.

cpaulgilman:
* Remove automatic weather file updates to avoid compatibility issues with newer versions of NSRDB API.

* Check for config changes from Flat Plate PV with battery enable to PV Battery for all version numbers newer than 2015.1.15. Function for 2020.2.29 changes config before older functions are called.


cpaulgilman:
Fixes #518

deploy/runtime/versions.lk
```diff
@@ -991,8 +881,13 @@ version_upgrade{'2015.8.17'} = define() {
 	}
 	
 	// Add two new dispatch modes, pushed manual to index 2002
-	if( cfg.tech == 'Flat Plate PV' && (cfg.fin == 'Residential' || cfg.fin == 'Commercial' ) && value('en_batt') == 1)
-		value("batt_dispatch_choice",2, "Keep as manual dispatch");
+	if( ( cfg.tech == 'Flat Plate PV' && value('en_batt') == 1 )  || cfg.tech == 'PV Battery') // in case config was changed from flat plate pv to pv battery in 2020.2.29 function
+	{
```
sjanzou, line 885: Good catch!

sjanzou (approved):
I restarted the Travis build - there was an error before...
Thanks for digging through this!

deploy/runtime/versions.lk
```diff
@@ -902,10 +789,13 @@ version_upgrade{'2015.11.16'} = define() {
...
 	const cfg = config();
-	if( cfg.tech == 'Flat Plate PV' && (cfg.fin == 'Residential' || cfg.fin == 'Commercial' ) && value('en_batt') == 1)
+	if( ( cfg.tech == 'Flat Plate PV' && value('en_batt') == 1 ) || cfg.tech == 'PV Battery') // in case config was changed from flat plate pv to pv battery in 2020.2.29 function
 	{
-		x = oldvalue('batt_replacement_cost');
-		value( 'batt_replacement_cost', x[0] ); // batt_replacement_cost is now single value
+		if ( cfg.fin == 'Residential' || cfg.fin == 'Commercial' ) 
+		{
+			x = oldvalue('batt_replacement_cost');
+			value( 'batt_replacement_cost', x[0] ); // batt_replacement_cost is now single value
```
brtietz, line 797: It looks like we reversed ourselves on this between 2020.2.29 and 2020.11.29 - the battery replacement cost is an array again. I assume we should leave this as is for the intermediate functions?

brtietz (approved):
Code looks good overall - I had one question about battery replacements. I merged patch into this and pushed it again since it looks like Travis was having trouble getting a pass.

deploy/runtime/versions.lk
sjanzou: My understanding is that batt_replacement_cost has always been an array
image

I cannot find 'batt_replacement_cost' in any UI forms on any compute modules - the variable used is 'om_replacement_cost1':
image

brtietz: Ah right, I was thinking of batt_cycle_cost which I changed in the latest version. It looks like the transition from batt_replacement_cost to om_replacement_cost1 happened in the upgrade to 2020.2.29.

cpaulgilman: batt_replacement_cost was originally a number on the Battery page, then we changed it to an array and moved it to the O&M page. Here's a screenshot from SAM 2016.3.14:

image

PR #529, opened 2021-02-05
Lcos cost inputs

mjprilliman:
-New PV+Battery OM cost UI form with updated variable names for financial models
-New storage salvage value input that still needs to be workshopped 
-Labels still up for changing for OM cost definitions based on industry standards
-Changes to cashflow to include LCOS cash flows


deploy/runtime/defaults/Flat Plate PV_Residential.txt
```diff
@@ -391,6 +391,18 @@ batt_max_power
...
 1
 0
+batt_salvage_percentage
+1
+1
+1
+1
+0
+batt_salvage_value
+1
```
sjanzou, line 401: Why are the other defaults changing? I get the ones you added but did the module and inverter change?

sjanzou (changes requested):
The tests are failing - most likely due to the default changes. Can you run all the default configurations and check test_results_win64.csv for valid results and then commit and push those results for Travis to match?

cpaulgilman (changes requested):
Layout of new O&M cost inputs looks nice.

Be sure to add/update UI group name so inputs appear in groups with other related variables in lists for LK script editor, parametrics, etc.

image

Also, I think I would prefer these UI forms to be named "O and M Costs PV", "O and M Costs Fuel Cell", etc. so they appear together in the IDE to make them easier to find when working on these O&M input variables.

cpaulgilman (commented):
I'm getting these messages when I create a new PV Battery / Single Owner case:

```
No default value found for 'wf_nrecords' in external file (PV Battery/Single Owner), using internal default
No default value found for 'vdcmax_inverter' in external file (PV Battery/Single Owner), using internal default
```

```
Defaults file is likely out of date: PV Battery_Single Owner.txt

Variables: 1 loaded but not in configuration, 0 wrong type, defaults file has 1130, config has 1131

Would you like to update the defaults with the current values right now?
(Otherwise press Shift-F10 later)

Loaded but don't exist in config: ui_batt_is_dc
```

cpaulgilman (approved):

PR #536, opened 2021-02-12
Make ppa and lease term variables in TPO model parametric

mjprilliman:
-First year monthly price and first year ppa price and their respective escalation rates now parametric


mjprilliman:
[TPO_parametric.zip](https://github.com/NREL/SAM/files/5973655/TPO_parametric.zip)

cpaulgilman (approved):
Looks good. Might consider making `lease_or_ppa` available as a parametric as well as we've done for the PPA models solution mode:

image

It's potentially confusing to run a parametric on monthly lease price when in PPA price mode, so this would at least make it possible to set the mode from parametrics.

Thanks,
Paul.

janinefreeman (approved):

PR #540, opened 2021-02-19
Capex table inputs

mjprilliman:
-Option for PV capital costs to be input as function of interpolation tables for DC, AC, land costs. 
-Reorganized existing system costs to move sales tax outside of indirect costs
-Hide sales tax parameter on financial parameters page when capex table input mode is selected


cpaulgilman (approved):
This looks good for the Beta.

For the final version, we should think about default values so that the red dot is visible for all default configurations. Maybe two sets of defaults, one for kW scale (residential, commercial, third party) and one for MW scale (PPA, merchant)?

mjprilliman:
> This looks good for the Beta.
> 
> For the final version, we should think about default values so that the red dot is visible for all default configurations. Maybe two sets of defaults, one for kW scale (residential, commercial, third party) and one for MW scale (PPA, merchant)?

I agree. I tried to use PV benchmark reports for a one-size-fits-all default approach but that does not look visually appealing

PR #541, opened 2021-02-19
Fix NSRDB API error handling

### LINKED ISSUE #531: NSRDB download intermittent error
**Describe the bug**

When NSRDB API service is down, attempt to download a weather file from Solar Resource page results in:

```
Could not evaluate callback function:btn_download->on_change
[507]: error: _s
[507]: failed to evaluate function call argument _d to '_s()'
[507]: failed to evaluate function call argument _d to '_s()'
[507]: eval error in statement list
[497]: eval error in statement list
[468]: eval error in statement list
[468]: error inside function call
[644]: eval error in statement list
[609]: eval error in statement list
[609]: error inside function call
[115]: eval error in statement list
```

**To Reproduce**

1. Modify UI callback in Solar Resource Data UI form to force error from API:

image

2. Try to download a weather file.

**Expected behavior**

Callback should gracefully handle API errors with a useful message.

cpaulgilman:
Fix #531

### DIFF
--- a/deploy/runtime/webapis.conf
+++ b/deploy/runtime/webapis.conf
@@ -4,7 +4,7 @@
 android_build=https://sam.nrel.gov/sites/default/files/content/mobile/android/readme.html
 forum=https://sam.nrel.gov/support
 ios_build=https://sam.nrel.gov/sites/default/files/content/mobile/ios/readme.html
-nsrdb_query=https://developer.nrel.gov/api/nsrdb/v2/solar/nsrdb-data-query.json?&api_key=<SAMAPIKEY>&email=<USEREMAIL>&wkt=POINT(<LON>%20<LAT>)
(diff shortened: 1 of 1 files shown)


deploy/runtime/webapis.conf
```diff
@@ -4,7 +4,7 @@
 android_build=https://sam.nrel.gov/sites/default/files/content/mobile/android/readme.html
 forum=https://sam.nrel.gov/support
 ios_build=https://sam.nrel.gov/sites/default/files/content/mobile/ios/readme.html
```
sjanzou, line 6: Is this the same as PR#58 into SAM-private?

sjanzou (approved):
Just checking the webapis.conf was updated in SAM-private too...  I will double check when building the patch

deploy/runtime/webapis.conf
cpaulgilman: Yes, SAM-private PR#58 is supposed to go with this one.

PR #559, opened 2021-03-10
Added gen as time series default for cases

### LINKED ISSUE #513: For the graphics related to DView, I would like default selections in all cases so that the results do not open up blank.
**Is your feature request related to a problem? Please describe.**
The problem is that for new users and demonstrations, it's not clear what these capabilities will do and the number of choices and options (particularly all the various energy outputs for the heat map) are extensive and overwhelming. 

**Describe the solution you'd like**
For every graphic and every combination, select an appropriate "start" variable to display and have taht selected and updated when the user first goes there. For teh bulk of these, the default should be really clear (hourly AC energy out for example).

**Describe alternatives you've considered**
Planning how to select those for demonstration.

**Additional context**

mjprilliman:
-When no variables are selected in Time Series, Profiles, Heat map, PDF CDF results tabs, "gen" or System power generated (kW) is selected by default for technologies that output the "gen" variable
-Also modifies the Data tables update on simulation to select "gen" when available if no variables are selected
-Theoretically the default variables that appear could change depending on the case technology and financial
-Closes #513


sjanzou (approved):
Thanks Matt!

mjprilliman:
Closing as the root branches were messed up. The additions were added to patch in a new PR.

PR #565, opened 2021-03-19
Fix Stochastic UI Issues

### LINKED ISSUE #564: Stochastic simulations are tricky to use
There are several usability issues with the stochastic simulations page.

**1. Weather file in stochastic inputs table is different from file in inputs browser**

The weather file variable used in stochastic.cpp is file_name, which is different from solar_resource (current library selection) in the UI. A Location and Resource UI equation sets file_name to solar_resource_file, but solar_resource_file is both calculated and indicator, so it is not shown in the inputs browser. 

image

**Solution:** On the Location and Resource UI form, clear the Calculated and Indicator check boxes for solar_resource_file, and change the label for solar_resource from "Location" to "Solar resource library current selection".

**2. Computed samples do not match output table**

In order to get the Computed Samples table to match the output table, the user has to know to click Compute samples before clicking Run simulations. (I wasted a bunch of time comparing outputs to inputs for samples that I generated after running simulations.)

**Solution:** Show the generated samples table when the user clicks Run simulations. A better solution would be to show samples and outputs in a single table like parametric simulations, see #22.

Additional enhancements and fixes:

*  Change "Enable weather file analysis" check box label to "Include normal distribution of weather files based on DNI or GHI".

*  "Show CDF" button export to a table instead of message box.

cpaulgilman:
Fixes #564 -- See issue for what to test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,9 @@ before_script:
   - export SAMNTDIR=/home/travis/build/$user/SAM
   - if [[ "$TRAVIS_PULL_REQUEST_BRANCH" == "" ]]; then export BRANCH_NAME=$TRAVIS_BRANCH; else export BRANCH_NAME=$TRAVIS_PULL_REQUEST_BRANCH; fi
   - git clone --depth=1 https://github.com/google/googletest /home/travis/build/$user/googletest
-  - git clone --depth=1 --branch=$BRANCH_NAME https://github.com/$user/lk $LKDIR; if [[ $? != "0" ]]; then git clone --depth=1 --branch="develop" https://github.com/$user/lk $LKDIR; fi
-  - git clone --depth=1 --branch=$BRANCH_NAME https://github.com/$user/wex $WEXDIR; if [[ $? != "0" ]]; then git clone --depth=1 --branch="develop" https://github.com/$user/wex $WEXDIR; fi
-  - git clone --depth=1 --branch=$BRANCH_NAME https://github.com/$user/ssc $SSCDIR; if [[ $? != "0" ]]; then git clone --depth=1 --branch="develop" https://github.com/$user/ssc $SSCDIR; fi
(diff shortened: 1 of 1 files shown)


deploy/runtime/ui/Solar Resource Data.txt
```diff
@@ -2477,7 +2477,7 @@ TextEntry
 solar_resource
 3
 4
-Location
+Solar resource library current selection
```
sjanzou, line 2480: Do you want to change in this pull request?

sjanzou (commented):
Not sure if you wanted to change the Solar Resource Data UI form in this pull request. Great job with the tables!

deploy/runtime/ui/Solar Resource Data.txt
cpaulgilman: Yes. That addresses Item 1 in #564.

test_results_win64.csv
```diff
@@ -119,26 +119,26 @@ Generic System,All Equity Partnership Flip,756864000,32.812180,36.422478,;
...
 PV Battery,Commercial,911877.375000,4.274710,NA,; 
 PV Battery,Third Party,34803.324219,NA,NA,; 
 PV Battery,Host Developer,911877.375000,13.250628,10.808079,; 
-PV Battery,Single Owner,92853616,7.822251,9.770243,; 
-PV Battery,Merchant Plant,92853616,6.944926,NA,; 
-PV Battery,Leveraged Partnership Flip,92853616,7.926883,9.770243,; 
-PV Battery,All Equity Partnership Flip,92853616,9.339340,10.855825,; 
-PV Battery,Sale Leaseback,92853616,9.509975,9.770243,; 
-PVWatts Battery,Residential,8022.540527,11.707731,NA,; 
+PV Battery,Single Owner,92853624,7.822259,9.770245,;
```
sjanzou, line 126: Not sure why these are changing... do the results need to be updated?

sjanzou (changes requested):
Can we leave the test_results_win64.csv out?

PR #566, opened 2021-03-19
Pysam v2.2.1

dguittet:
## Version 2.2.1, Mar 19, 2021 ~ SAM 2020.11.29 Revision 1, SSC Version 253
* Add `unassign` function to modules (NREL/pysam#65)
* Fix mypy errors in stubs (NREL/pysam#47)
* Fix BatteryStateful bugs: memory leak; unintialized pointer in some constructors; fix `setup` error messaging (NREL/pysam#67)
* BatteryStateful features: ability to use adaptive timestep in simulation; speed up execution; Smith 2017 Life Model
* Voltage cutoff feature for battery model (Battery & BatteryStateful)
* Battery sizing tool can use per-module scaling for surface area (NREL/pysam#77)
* Updates to ResourceTools for weather download (NREL/pysam#53, NREL/pysam#50)
* Leap year 2/29 date is now acceptable
* Fix docs for TcsMoltenSalt (NREL/pysam#66)
* Update Import from SAM GUI example to use "PySAM JSON" option in Generate Code
* Better error messaging (NREL/pysam#74)
* [SAM Release updates for Version 2020.11.29 Revision 1](https://nrel.github.io/SAM/doc/releasenotes.html)

### DIFF
--- a/api/api_autogen/builder_PySAM.cpp
+++ b/api/api_autogen/builder_PySAM.cpp
@@ -1062,6 +1062,8 @@ void builder_PySAM::create_PySAM_files(const std::string &cmod, const std::strin
 
         for (const auto& it : vardefs) {
             std::string var_symbol = it.first;
+            if (var_symbol == "global")
+                continue;
             fx_file << "\t" << var_symbol << " = " << statictype_str[it.second.type_n] << "\n";
         }
         fx_file << "\n\n";
--- a/api/include/SAM_TcsgenericSolar.h
+++ b/api/include/SAM_TcsgenericSolar.h
@@ -733,7 +733,7 @@ extern "C"
 
 	SAM_EXPORT double* SAM_TcsgenericSolar_Outputs_gen_aget(SAM_table ptr, int* length, SAM_error *err);
 
-	SAM_EXPORT double* SAM_TcsgenericSolar_Outputs_glob_aget(SAM_table ptr, int* length, SAM_error *err);
+	SAM_EXPORT double* SAM_TcsgenericSolar_Outputs_global_aget(SAM_table ptr, int* length, SAM_error *err);
 
 	SAM_EXPORT double* SAM_TcsgenericSolar_Outputs_hour_aget(SAM_table ptr, int* length, SAM_error *err);
 
--- a/api/modules/SAM_TcsgenericSolar.cpp
+++ b/api/modules/SAM_TcsgenericSolar.cpp
@@ -1368,12 +1368,12 @@ SAM_EXPORT double* SAM_TcsgenericSolar_Outputs_gen_aget(SAM_table ptr, int* leng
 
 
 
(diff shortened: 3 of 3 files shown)


api/api_autogen/library/defaults/Battery_GenericBatteryMerchantPlant.json
```diff
@@ -37,7 +35,7 @@
 			"batt_replacement_schedule_percent": [0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000],
 			"batt_surface_area": 89.0141,
 			"en_batt": 1,
-			"om_replacement_cost1": [500.000000]
```
brtietz, line 40: What happened such that these variables are just now getting updated? I would have expected this change in 2.2.0.

api/api_autogen/library/defaults/Pvsamv1_FlatPlatePVLCOECalculator.json
```diff
@@ -182,24 +182,24 @@
 			"cec_v_oc_ref": 64.4
 		},
 		"CECPerformanceModelWithUserEnteredSpecifications": {
-			"6par_aisc": 0.004,
```
brtietz, line 185: It looks like these are still 6par instead of sixpar on develop (and pysam-v2.2.1) in ssc. Is this going to break anything?

brtietz (changes requested):
I'm missing the function that translates "sixpar" into "6par" for ssc. No objections to the name change, I just want to make sure these new JSON files will work.

Other than that looks good.

api/api_autogen/library/defaults/Pvsamv1_FlatPlatePVLCOECalculator.json
dguittet: No it still works because the `marshal` module in Python automatically converts it to "six" when creating the default `.df` files

api/api_autogen/library/defaults/Battery_GenericBatteryMerchantPlant.json
dguittet: Not really sure why. 2.2.0 was built shortly after the release.

brtietz (approved):

api/api_autogen/library/defaults/Pvsamv1_FlatPlatePVLCOECalculator.json
dguittet: Nevermind, I misinterpreted this. The problem is that the `6par_aisc` name as a pysam variable never worked. In PySAM we have to rename it so that it doesn't start with a number. In SSC, we could change everything, but that would be really messy. The part that we could do better is in the PySAM JSON Code Generator.

PR #576, opened 2021-04-01
Restore NMC battery defaults for lifetime

### LINKED ISSUE #574: Switching to NMC battery chemistry on develop doesn't work
**Describe the bug**
When using the "battery type" menu on develop, I get the following error:

```
Could not evaluate callback function:batt_type->on_change
[242]: error in call to 'value()': data type mismatch attempting to set 'batt_lifetime_matrix' (matrix) to <null> (null)[207]: eval error in statement list
[207]: error inside function call
[198]: eval error in statement list
[198]: error inside function call
[355]: eval error in statement list
[355]: error inside function call
[138]: eval error in statement list
[135]: eval error in statement list
[135]: error inside function call
[13]: eval error in statement list
```

**To Reproduce**
1. Go to the battery system page
2. Click on "battery type: lithium ion nmc" in the drop down.
3. Say you want to replace the default values

**Expected behavior**
Default value replacement should happen as on patch.

**Additional context**
Not an issue in 2020.11.29 r1. I haven't tested patch.

dguittet:
fix #574

Restore the old default behavior of using the lifetime cycle table and calendar model. Need to change SAM GUI to use NMC analytic model from Smith 2017 as default. Then these old defaults can be removed.

### DIFF
--- a/deploy/libraries/BatteryStateful/BatteryStateful_NMCGraphite.json
+++ b/deploy/libraries/BatteryStateful/BatteryStateful_NMCGraphite.json
@@ -1,19 +1,19 @@
 {
-	"defaults": {
-		"ParamsCell": {
-			"C_rate": 0.2,
-			"Qexp": 2.584,
-			"Qfull": 3.2,
-			"Qnom": 3.126,
-			"Vexp": 3.53,
-			"Vfull": 4.2,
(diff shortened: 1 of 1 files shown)


brtietz (approved):
This fixed the bug! Thanks for the quick fix!

PR #581, opened 2021-04-06
Wind resource distribution

### LINKED ISSUE #533: Wind uncertainty language should indicate you can likely ignore the inputs.
I must have been daydreaming when we talked about Wind Uncertainties. I was tracking down something else and ran across the Uncertainties page on the Wind model. Coming in cold, I don’t think that the text at the top of the page really explained what the hell all of this is. The help was more helpful with the statement below. I might propose adding something else to the top of the page (just before the sentence “All uncertainty components and subcomponents…” such as “This page contains the inputs that calculate probability distributions displayed on the Uncertainties Tab of the results page.  If you are not interested in viewing uncertainties, you may ignore these inputs.”  Otherwise, it’s a pretty scary page.

The Uncertainties inputs allow you to specify uncertainties for different input parameters to calculate the probability distributions displayed on the Uncertainties tab of the Results page.
The uncertainty model is draws from the IEC 61400-15 uncertainty framework as described in Damiani, R. (2018). Uncertainty and Risk Assessment in the Design Process for Wind. National Renewable Energy Laboratory, NREL/TP-5000-67499 (PDF 8.9 MB).

### LINKED ISSUE #547: Improve wind resource option labels, documentation and defaults
* Add description of wind resource distribution table option to Help.

* Better defaults for speed/direction distribution table to illustrate how the table works.

* Rename options for clarity:
 
image

Name options (note that distribution table is for both speed an direction):

Choose Wind Resource File
Define Wind Speed Weibull Distribution
Define Wind Resource Distribution
 
Other options that are consistent with the other names might be:
 
Wind Speed Distribution Table
Wind Resource Distribution Table
Wind Speed Probability Table

Also note related issue on Wind UI forms: https://github.com/NREL/SAM/issues/533

mjprilliman:
-Turns off wind resource uncertainty by default
-Hides Uncertainties results tab when uncertainty is not specified
-More meaningful defaults
-Added system capacity to wind power metrics table
-Closes #533 
-Closes #547


cpaulgilman (approved):
Looks good. I noticed that the Year 1 heat map on the Summary tab is empty, but perhaps that -- will resolve itself when this gets merged into Patch.

mjprilliman:
> Looks good. I noticed that the Year 1 heat map on the Summary tab is empty, but perhaps that -- will resolve itself when this gets merged into Patch.

What were the steps you used to get that behavior? The autographs script is supposed to only produce a heatmap for the wind model in the case where the wind resource file is specified, but it should never be empty if patch on ssc is up to date.

cpaulgilman:
@mjprilliman, I tested with all of the possible wind resource options and the Summary tab heat map is working properly. I did not see an empty graph in these most recent tests.

cpaulgilman:
@mjprilliman, I think this can be closed.

PR #592, opened 2021-04-22
Show correct weather file in Wind PDF report, fixes #591

### LINKED ISSUE #591: Wind Power PDF Report Shows Incorrect Weather File
**Describe the bug**
The PDF report shows the weather file for the current selection in the wind resource library even when **Use a wind resource file stored on your computer** is checked. 

**Expected behavior**
Report should show weather file used for simulation.

cpaulgilman:


mjprilliman (approved):
Looks good to me.

PR #593, opened 2021-04-22
Batt chem defs

dguittet:
Update the lead acid voltage defaults so that the voltage table reflects the Vnom_default = 2

Fixes part of NREL/ssc#565


brtietz:
Shoot - I had tagged the issue with patch 2. Would you be willing to cherry-pick the commits to a branch based on patch so we can have the fix there?

dguittet:
absolutely! good catch. restarting PR

PR #595, opened 2021-04-22
fix lead acid voltage table's voltage to be same as Vnom_default

dguittet:
Update the lead acid voltage defaults so that the voltage table reflects the Vnom_default = 2

Fixes part of NREL/ssc#565


dguittet:
Not sure why the sam travis tests are failing due to changed in the physical trough-- these changes here only affect the leadacid battery's defaults. All the battery configurations use li-ion battery as default so there shouldn't be any effect on the sam tests.

brtietz:
Ty just fixed that this morning: https://github.com/NREL/SAM/commit/9529aacc21a3f1cef4afe69492ab022dd786e366

Up to you whether to merge patch into this branch now or after the test passes.

dguittet:
Thanks! Just merged and will wait for the test to pass.

PR #600, opened 2021-04-30
Remove tower heliostat availability variable to fix #50

### LINKED ISSUE #50: Power tower heliostat availability not used in simulations
Heliostat availability is used in design calculations but not during performance simulations. Workaround solution is to use the Constant loss in the Edit Losses window to represent heliostat availability.

cpaulgilman:
* Remove helio_active_fraction widgets from Tower Solar Pilot Solar Field and Tower Solar Field UI forms, but leave variable as calculated value because variable is required by TCS.

* Change label to "Heliostat availability not used" to minimize confusion for LK scripting.

* Set default helio_active_fraction value to 1 for MSPT and DSPT defaults, and also in UI variable default value. Note that DSPT is a "retired" configuration.

* Set default value of sf_adjust:constant to 1 for MSPT configurations to account for 99% value of old variable.

* Update test results file -- change in constant loss to replace heliostat availability may result in small differences in MSPT results.

* Update Help (will commit separately before release).


cpaulgilman:
This issue needs more consideration (see #50 discussion), so closing pull request without merging.

PR #603, opened 2021-05-01
FOM Battery Remove Sell Rate

cpaulgilman:
* For Battery FOM, change label for Electricity Rates page to Electricity Purchases and revise descriptions in UI labels. Add new Help page and revise descriptions.

* Correct conditional display of electricity rates inputs: No sell rates, enforce buy all / sell all metering option.

* Fix typo in Group Name for battery dispatch charge limits.

* UI callback function to display a message and set sell rate to zero if user opens a file created with SAM 2020.11.29 r1 with non-zero sell rates or if LK script sets sell rate to non-zero value.

* Modify version upgrade script to handle opening old FOM battery cases where sell rate is non-zero. To test, try opening these files created with SAM 2020.2.29 or older that incorrectly allowed TOU sell rates: 
[fom-battery-sell-rate-tests.zip](https://github.com/NREL/SAM/files/6410391/fom-battery-sell-rate-tests.zip)

### DIFF
--- a/deploy/runtime/defaults/Fuel Cell_Single Owner.txt
+++ b/deploy/runtime/defaults/Fuel Cell_Single Owner.txt
@@ -4570,7 +4570,7 @@ ur_metering_option
 1
 1
 1
-0
+4
 ur_monthly_fixed_charge
 1
 1


brtietz:
This is great! I particularly like that this page now makes it clear that PPA is the alternative to the retail rate. Three testing notes/suggestions:

(1) Change the default for "compensation rate for net excess generation" to $0:

image

I feel like having a non-zero number in a greyed out box is confusing.

(2) The Fuel Cell page still allows the full suite of distributed generation options for Single Owner, these should line up with other FOM configurations:

image

(3) These constraints are not enforced with the "save / load rate from file" dialog. I can force a change to net billing using this file: 
[test_load_net_billing.zip](https://github.com/NREL/SAM/files/6415882/test_load_net_billing.zip) Is this something we want to allow?

brtietz (changes requested):
See previous comment for requested changes (probably should have just written them up here).

cpaulgilman:
@brtietz Thanks for the feedback. I made changes to address the two issues you noted. I also ran the test script again in case changing the default excess compensation rate to zero affected results. The default rate was non-zero for several configurations, including ones like Physical Trough Commercial, and I set them all to zero.

brtietz:
The fixes for (1) and (3) look great!

For (2) it looks like the default for ur_metering_option for fuel cell single owner needs to change, now it's stuck on net metering:

image

cpaulgilman:
My latest commit https://github.com/NREL/SAM/pull/603/commits/7fab175c476913c6678ff8f50954d22bdcbf258b fixes an issue when importing a rate data CSV with energy charge max usage units value of -1, which can happen when importing a CSV file containing data for a rate from the URDB where max usage units is set to -1.

It also modifies the URDB download process to change usage units = -1 to zero to avoid this issue to begin with.

brtietz (approved):
Recent changes look great, thanks!

PR #606, opened 2021-05-03
Changed duration sizing defaults to kwh option, changed variable

mjprilliman:
-Fixes lost value issue when changing between kwh and h battery sizing options
-Implements bidirectional overwrites when switching between options to keep values for each option unless manually changed
-Sets default to kwh option for all battery configurations, as this is the existing method for sizing the battery.


brtietz (changes requested):
The functions for switching between kWh and h are working great, but my saved files from 2020.11.29r1 are still having their durations overwritten to a two hr battery. It's just displayed in kWh now.

brtietz (approved):
Looks good as of the latest commit. Thanks!

sjanzou:
@mjprilliman, pull request looks good

cpaulgilman (approved):

PR #607, opened 2021-05-03
Migrate trough UI eqns

Matthew-Boyd:


sjanzou:
@Matthew-Boyd, looks like your branch was made off of develop but has a Patch 2 Milestone associate with it. Probably no way to merge into the patch branch if originally from develop. Also, Travis is failing due to default results changes. Can you check them locally?

Matthew-Boyd:
@sjanzou, Thanks Steve. I used the wrong branch, meant to submit my develop
branch. Travis might still fail though, so I'll work to ensure it doesn't.

PR #609, opened 2021-05-06
Fix #605 NSRDB time zone issue

### LINKED ISSUE #605: NSRDB downloads from himawari and msg-iodc set time zone to 0
**Describe the bug**
Weather files from the NSRDB himawari and msg-iodc datasets return time zone = 0 instead of offset from UTC.

**To Reproduce**
1. On Location and Resource page, click Advanced Download and download a weather file for 5.28,115.23 (himawari) or 31.9,35.4 (msg-iodc).
2. Note the time zone under *Weather Data Information*.

**Workaround**
Edit the weather file after it downloads and change the time zone field to the correct value.

**Expected behavior**
Time zone should be local time zone. 

**Additional context**
Note utc parameter in URL in sam_nsrdb_download_log.txt is set to false. Need to verify for other NSRDB datasets.

cpaulgilman:
* Add explicit "&utc=false" attribute to all NSRDB calls to ensure local time option is enabled.

* Add attribute list for himawari endpoint.

* Add clear sky DNI, DHI, GHI to list of attributes to partially address https://github.com/NREL/ssc/issues/151.

* Clean up message text in Solar Resource Data UI callback when no file is available from callback download attempt.

### DIFF
--- a/deploy/runtime/ui/Solar Resource Data.txt
+++ b/deploy/runtime/ui/Solar Resource Data.txt
@@ -2624,7 +2624,7 @@ equations{'ui_step_minutes'} = define() {
 	return(${step}/60);
 };
 
-32933
+32888
 on_load{'Solar Resource Data'} = define() {
 	//msgbox('on load');
 	//hide legacy data for 2020.2.29 Patch 3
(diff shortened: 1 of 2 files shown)


sjanzou (approved):
Hopefully, utc=false will not change any time soon!

PR #612, opened 2021-05-10
Replaces pull request SAM 608 used with ssc pull request 574

sjanzou:


Matthew-Boyd (approved):
Pulled and tested--runs good for the trough models.

dguittet (approved):
Thanks @sjanzou for sorting that out!

PR #613, opened 2021-05-10
Fixed DView functions relying on automatic selection of Gen

mjprilliman:
-Data tables Clear all button now correctly clear "gen" when pressed
-Profiles automatically shows all 12 month and annual plots for "gen" when no variables or only "gen" are selected

### DIFF
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -3388,10 +3388,12 @@ void fcall_rescanlibrary( lk::invoke_t &cxt )
 
 	if ( reloaded != 0 )
 	{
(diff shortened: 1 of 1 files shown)


cpaulgilman:
I am able to crash SAM reliably following these steps:

1. Create a PV Battery / Residential case.
2. Run simulation.
3. On the Data Tables tab, under Monthly Data, add a few variables. I tested with Energy charge with system, Energy charge with system before credits, Energy charge without system, Energy charge to battery from grid.
4. Run another simulation.
5. On Data Tables tab, switch to Lifetime Hourly Data
6. Go to Profiles tab. SAM crashes.

You might need to repeat 4-6 a couple of times to replicate the crash.

PR #621, opened 2021-05-19
Reorder code generator to set all inputs and then call all compute modules at end fixes #614

### LINKED ISSUE #614: Change order of variable definition and cmod calls in LK generator
Current LK script generator hardcodes input values for the first cmod, then calls the cmod, then hardcodes input values for the next cmod, etc. This causes an issue when the first cmod resets the value of an input value for subsequent cmods. In this case, after the first cmod resets the input value, it is overwritten by the hardcoded value in the lk script before the next cmod.

Proposed solution: hardcode all input values before any cmod are called. Then, if cmod resets an input value, the new value will carry through to the next cmod.

sjanzou:
Addresses SAM #614

### DIFF
--- a/src/codegenerator.cpp
+++ b/src/codegenerator.cpp
@@ -485,6 +485,7 @@ bool CodeGen_Base::GenerateCode(const int &array_matrix_threshold)
 
 	// run compute modules in sequence (INOUT variables will be updated)
 //	for (size_t kk = 0; kk < simlist.size(); kk++)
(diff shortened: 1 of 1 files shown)


sjanzou:
@tyneises, @Matthew-Boyd please build using develop branches or children of develop branches of lk, wex, and ssc and the SAM_614 branch of SAM and test for desired functionality.

tyneises:
Functionality looks good to me

tyneises (approved):

Matthew-Boyd (approved):

PR #624, opened 2021-05-21
Parametrics updates from EPRI CRADA work

sjanzou:


src/graph.cpp
```diff
@@ -130,7 +134,7 @@ bool Graph::Read( wxInputStream &is )
 	wxDataInputStream ds(is);
 
 	unsigned short identifier = ds.Read16();
-	/*unsigned char ver = */ ds.Read8(); // read the version number, not currently used...
+	unsigned char ver = ds.Read8(); // read the version number, not currently used...
```
brtietz, line 137: Please update the comment, as this is now used.

src/graph.cpp
```diff
@@ -531,6 +543,257 @@ static const char *s_monthNames[12] = { "Jan", "Feb", "Mar", "Apr", "May", "Jun"
...
+			GetYAxis1()->SetLabel(m_g.YLabel);
+			GetXAxis1()->SetLabel(m_g.XLabel);
+			
+		}
+		ndata = sims.size();
+		return 0;
+
+	}
+	else
+	{ // TODO ensure m_g.X.count == 1 m_g.Y.count == 1
```
brtietz, line 655: Is this still relevant? It looks like vv->Length() is acceptable later.

src/main.cpp
```diff
@@ -117,6 +117,10 @@ static SamApp::ver releases[] = {
 //intermediate version numbers are required in this list in order for the version upgrade script (versions.lk) to work correctly
 //please clarify the reason for the new version in a comment. Examples: public release, variable changes, internal release, public beta release, etc.
 //the top version should always be the current working version
+		{ 2021, 4, 19 }, // 2021.4.19 ssc 255 EPRI Beta expires 4/19/2022
```
brtietz, line 120: Most recent version should be for the patch, correct? It looks like develop needs to be merged into this branch. I wonder if a command-line merge would solve the Travis issues as well.

src/parametric.cpp
```diff
@@ -963,7 +1209,7 @@ wxArrayString ParametricViewer::getFromExcel(const wxString& input_name, int& ro
 #endif // __WXMSW__
 }
 
-void ParametricViewer::SaveToCSV()
+void ParametricViewer::SaveToCSVOld()
```
brtietz, line 1212: It looks like this isn't used anywhere now. I think my first choice would be to remove it, but if you feel strongly about keeping it, can you add a comment explaining the differences with the current function.

src/parametric.cpp
```diff
@@ -984,7 +1230,7 @@ void ParametricViewer::SaveToCSV()
 
 }
 
-void ParametricViewer::SendToExcel()
+void ParametricViewer::SendToExcelOld()
```
brtietz, line 1233: Same note as SaveToCSVOld

brtietz (changes requested):
The functionality in SAM looks good, but I saw some need for additional comments and clean up in the code. Is the Travis issue a Linux issue or something with the Travis config?

src/graph.cpp
sjanzou: The one and one check is the check to plot output variable versus input variable for single outputs that are not monthly or timeseries based. An artifact of the GraphCtrl class in SAM that is a wrapper with limited functionality of PlotCtrl widget in wex so that autograph functions in the autograph.lk script can be used to show graphs on the results page. An opportunity to upgrade or expand some basic functionality in SAM.

src/graph.cpp
sjanzou: updated to version 2

src/main.cpp
sjanzou: This branch was made off of develop. And is kept up to date with develop. This includes the latest beta builds. Patch does not include them. Darice usually merges patch into develop after a PySAM update.

sjanzou (commented):
All issues addressed and checks passed

cpaulgilman:
When filtering columns, the graphs clear and attempt to update based on filtered value. Graphs should not be affected by how the list is filtered.

I'm not sure we need both the right-click column fill functions and Quick Setup. The fill functions seem to replicate the functionality of Quick Setup. I would recommend limiting the right-click shortcut functions to filtering and sorting the table, and adding any functions for filling tables to Quick Setup.

There are some issues with how the fill functions work:

* Fill functions seem to work only after clicking another column in the table. I would expect them to work after clicking the command in the shortcut window.

* After using the Fill Down Sequence option, when I type a value in the column it automatically changes. I expect fill functions to stop after I leave the shortcut menu.

* "Fill down evenly" is not clear: Does it fill cells between the first and last rows?

mjprilliman (commented):
I mostly just looked at the UI which looks really good!

cpaulgilman (approved):
Ok to close this pull request and merge now to avoid potential conflicts with Develop. We'll revisit question of fill right-click functionality vs Quick Setup as a separate issue.

src/parametric.cpp
sjanzou: All old functions removed in the last commit. Thanks for the detailed review

PR #628, opened 2021-05-27
Battery Dispatch user interface reorganization

mjprilliman:
-Radio choice options to show/hide battery dispatch inputs depending on the dispatch model
-Different configurations for Battery BTM, Battery FOM, Fuel Cell BTM, Fuel Cell FOM
-New lk addpage function option: 'exclusive_hide' to hide exclusive_var tabs or dropdown


cpaulgilman:
@mjprilliman Nice work getting this working! I cleaned up the layout and conditional logic on the forms and fixed some issues with defaults. I think a lot of the code in the UI callbacks on some of the new forms is not needed, perhaps copied over when the forms were created, and should be removed. I can do that and do more defaults testing in a separate commit.

PR #641, opened 2021-06-29
add run_resilency_calcs to gui and defaults

brtietz:
Pairs with ssc PR https://github.com/NREL/ssc/pull/586

If new box is checked we get the old behavior when a critical load is specified. If the box is unchecked, the critical load can be specified without the resiliency runner running.

Want to merge this sooner while other work is ongoing to avoid conflicts in the defaults files. Wordsmithing on both PRs is welcome.

Do we want to specify a grid_outage default, or is a zero length array ok for that input (meaning no outages)?

### DIFF
--- a/deploy/runtime/defaults/Generic Battery_Commercial.txt
+++ b/deploy/runtime/defaults/Generic Battery_Commercial.txt
@@ -1,5 +1,5 @@
 1
-471
+472
 LeadAcid_q10
 1
 1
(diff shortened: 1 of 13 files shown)


cpaulgilman (approved):
This looks fine for now. There is no documentation of critical loads or resiliency in Help. I will write documentation and probably tweak UI labels after this is merged into Develop.

PR #643, opened 2021-06-30
Marine wave resource time series data

mjprilliman:
-New UI interface with time series data
-Significant_wave_height, energy_period arrays used to calculate wave energy converter performance
-API calls hidden, still in development
-New exclusive_var used to specify whether probability distribution or time series data are used in calculations
-New 3-hour time series outputs
-New wave resource time series data library with defaults matching locations of JPD file locations
-New function in library.cpp for handling changes to time series library


mjprilliman:
Renamed to match ssc branch and moved to new PR

PR #644, opened 2021-06-30
Fixed batt dispatch loaded page to Manual dispatch for FOM systems

mjprilliman:
-Fixes page misalignment with radio choice selection on new case load-in for FOM battery systems
-Fixed fuel cell dispatch behavior where inputs were visible and enabled even when en_batt == 0


cpaulgilman (changes requested):
For the Fuel Cell "Dispatch" page, can you move the FOM/BTM Storage Dispatch Options UI form directly below the Fuel Cell Dispatch form so that the battery forms are below the options? That will avoid the options jumping around as you change them.

There are large blocks of commented-out code in the UI callbacks for these forms. Can they be deleted?

Battery Dispatch Front of Meter
Battery Dispatch Manual
Battery Dispatch Options BTM

mjprilliman:
So I've been confused with these pages as even when the Manual Dispatch choice isn't selected for the battery, the fuel cell dispatch is still handled from this page so it must still be visible. So, moving the manual dispatch form to the bottom of the page would require some sort of exclusive_footer tag to be developed similar to exclusive_header as the page would always be underneath switching battery dispatch cases.

cpaulgilman:
Right. That is confusing. I think I like it better now with the fuel cell operation and storage dispatch options at the top. We can improve on this later if we have a fuel cell project to support that work.

cpaulgilman (approved):

PR #645, opened 2021-07-01
Wave model time series

mjprilliman:
-New UI interface with time series data
-Significant_wave_height, energy_period arrays used to calculate wave energy converter performance
-API calls hidden, still in development
-New exclusive_var used to specify whether probability distribution or time series data are used in calculations
-New 3-hour time series outputs
-New wave resource time series data library with defaults matching locations of JPD file locations
-New function in library.cpp for handling changes to time series library


cpaulgilman (approved):
Need to delete file_name from defaults for ME Wave / None configuration.

Do we need both wave_data_file_ts and wave_resource_filename_ts for the time series option? Same question for wave_data_file_freq and wave_resource_filename for the frequency option.

mjprilliman:
> Need to delete file_name from defaults for ME Wave / None configuration.
> 
> Do we need both wave_data_file_ts and wave_resource_filename_ts for the time series option? Same question for wave_data_file_freq and wave_resource_filename for the frequency option.

I suppose not. the vars with filename are what go into the compute modules. So I think I will keep those to be consistent.

PR #660, opened 2021-07-16
Rename ui_tech_capacity variables to om_tech_nameplate

mjprilliman:
-To maintain syntax of ui_ prefix only being used for UI specific variables
-All relevant financial models updated
-Updated defaults, OM cost forms


cpaulgilman (approved):
Looks good. I checked for default messages for all configurations and found none.

Issue #661, opened 2021-07-20
Battery Life Model Additions & Other Updates


dguittet:
In SSC there are two new battery life models for NMC Graphite https://github.com/NREL/ssc/pull/579 and LMO LTO https://github.com/NREL/ssc/pull/588 chemistries. These should be made available in SAM. Updates to LMO LTO defaults will be needed.

Also, battery thermal defaults for the size of the battery relative to the capacity should be updated for Commercial and utility-scale batteries. The current values are for residential-scale batteries and over-estimate the size of the battery for MW+ systems.

dguittet:
completed in #691

PR #662, opened 2021-07-21
Lmolto defs

dguittet:
A new LMO / LTO life model was added to SSC in https://github.com/NREL/ssc/pull/588. Here, add a BatteryStateful_LMOLTO.json file for updated LMO / LTO defaults that can be both loaded in SAM and used in PySAM. The values are taken from [Toshiba SCiB SPCS-LIB010A 10kWh](https://www.toshiba-itc.com/lithiumion/pdf/lithium_catalog.pdf).

In a future development, https://github.com/NREL/SAM/issues/661, the defaults can be better updated to use the new life model, but for now, just copy the existing life cycling matrix.


mjprilliman (approved):
LGTM

cpaulgilman (changes requested):
1. The lines starting at https://github.com/NREL/ssc/blob/fc993a6f8b5c195d2caddd723f60887f226be3f6/ssc/cmod_battery_stateful.cpp#L65 impose conditions on some battery input variables based on the value of `life_model`, but those conditions are not enforced in the user interface. Should battery lifetime inputs in the user interface be enabled/disabled based on the value of `life_model`, or do we want to allow the user to specify a combination of cycle degradation and calendar degradation even when `life_model` is not zero?

2. I think the Battery Lifetime UI form should indicate when one of the internal battery life models is enabled (i.e., when `life_model = 1` or `life_model = 2`) with a text description like "Li-ion-NMC life model enabled" or "Li-ion-LMO/LTO life model enabled".

3. In SSC, the units for the `life_model` input in the var table is "0/1" instead of "0/1/2".

dguittet:
Thanks @cpaulgilman for the comments.

1/2. The life_model options aren't available in the GUI yet. So in the GUI now, it should life_model should always be 0. When that's added, it will be made explicit. 

The confusing part about the defaults here are that they serve two sometimes disjunct purposes. Most defaults are shared by the GUI and by BatteryStateful, but when it comes to the life model, the GUI still uses the `cycle_matrix` and `calendar_choice` & other calendar inputs, whereas the BatteryStateful NMC & LMOLTO doesn't use any of those and only the `life_model`. So while each case uses different sets of inputs for the life modeling, both of them are in the default JSONs. Once #661 is done, we can move the GUI to use the life_model input and get rid of the unnecessary inputs.

Thanks for catching "0/1/2"! Fixing that.

PR #664, opened 2021-07-22
PV Smoothing to merge into develop to consolidate FOM options

sjanzou:


deploy/runtime/ui/Battery Dispatch PV Smoothing.txt
```diff
@@ -0,0 +1,1921 @@
...
+on_load{'Battery Dispatch PV Smoothing'} = define() {
+	on_change{'batt_dispatch_pvs_timestep_multiplier'}();
+	on_change{'batt_dispatch_pvs_ac_ub_enable'}();
+	on_change{'batt_dispatch_pvs_ac_lb_enable'}();
+	on_change{'batt_dispatch_pvs_short_forecast_enable'}();
+	value('batt_dispatch_choice', 5);
+	toggle_dispatch_pvs(); // do not enable unless mode == 5
+};
+
+function toggle_dispatch_pvs()
```
mjprilliman, line 1869: Change toggle_dispatch to toggle_dispatch_pvs as Manual Dispatch also uses toggle_dispatch()

deploy/runtime/ui/Battery Dispatch Automated FOM.txt
```diff
@@ -30,10 +30,11 @@ Selection
 0
 Items
 6
-3
+4
 Perfect look ahead
 One day look behind
 Look ahead to custom weather file
+PV smoothing
```
mjprilliman, line 37: Index Labels property on variables was overwriting Object properties Items list so PV Smoothing wasn't showing up. I committed the fix to this branch

test_results_win64.csv
```diff
@@ -1,145 +1,145 @@
...
-Generic Battery,Single Owner,73190888,20.643936,10.399951,; 
-Generic Battery,Merchant Plant,73190888,19.608706,NA,; 
-Generic Battery,Leveraged Partnership Flip,73190888,24.984268,28.888752,; 
-Generic Battery,All Equity Partnership Flip,73190888,28.859419,34.666504,; 
-Generic Battery,Sale Leaseback,73190888,29.022760,34.666504,; 
+PV Battery,Residential,8588.59,11.2055,NA,; 
+PV Battery,Commercial,911877,4.27471,NA,; 
+PV Battery,Third Party,34803.3,NA,NA,; 
+PV Battery,Host Developer,911877,13.2506,10.8081,; 
+PV Battery,Single Owner,996456,103.113,9.40234,; Notice: The simple efficiency and Sandia module models do not allow limiting module voltage to the MPPT tracking range of the inverter.
```
mjprilliman, line 127: Need to re-run Single Owner model as the defaults were all changed in the PVS debugging process.

brtietz:
Given that the weather file default in SAM is an hour I wonder if this red message should explicitly discuss a shorter timestep weather file:

image

mjprilliman (commented):
Fixed some issues from the integration of develop into PVSmoothing.

brtietz:
This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:

image

Do we want to provide the option for that configuration or remove it?

brtietz:
Looks like there are a few issues with generic battery FOM cases as well. This is with "import hourly or subhourly generation profile" selected:

image

brtietz (changes requested):
Detailed PV-Battery FOM configurations look good to me. Fuel Cell and Generic System cases have some bugs with the new PV Smoothing UI.

mjprilliman:
> This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:
> 
> image
> 
> Do we want to provide the option for that configuration or remove it?

Not sure on whether it should be included or not but this was a startup.lk error as the PV Smoothing form wasn't in the list for Fuel Cell FOM. I can push a fix if we do want to include this as an option

mjprilliman:
> Looks like there are a few issues with generic battery FOM cases as well. This is with "import hourly or subhourly generation profile" selected:
> 
> image

This stems from the ui_step_minutes calculated variable. The issue there is that when you select "import hourly or subhourly generation profile", The default for the energy_output_array variable has size 1. So is this a bug or a hint to the user that they actually need to input some sort of generation profile?

cpaulgilman (approved):
I did a little UI cleanup, including addressing @brtietz's comment about the warning message for subhourly weather data:

image

I think this option should only be available for PV Battery / Single Owner configurations for now. I could see an application for Generic Battery / Single Owner for users who want to run the model with measured output data, but this mostly seems to be a forecasting tool rather than a performance monitoring tool. For fuel cells would people expect an option to allow the fuel cell to contribute to PV ramp rate mitigation?

I pushed a change to handle that for now by displaying a UI message on Battery Dispatch Options FOM if `batt_dispatch_choice_ui = 3` when technology() != PV Battery, and come up with a better solution later.

cpaulgilman:
> 
> 
> > This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:
> > image
> > Do we want to provide the option for that configuration or remove it?
> 
> Not sure on whether it should be included or not but this was a startup.lk error as the PV Smoothing form wasn't in the list for Fuel Cell FOM. I can push a fix if we do want to include this as an option

@mjprilliman Can you push a fix to add the PV smoothing UI form to Fuel Cell configurations for now? After my latest commit, if you choose PV smoothing a red warning message should display saying that PV smoothing is for PV Battery only. Otherwise we will need a different version of batt_dispatch_choice_ui for PV and non-PV battery configs.

mjprilliman:
> > > This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:
> > > image
> > > Do we want to provide the option for that configuration or remove it?
> > 
> > 
> > Not sure on whether it should be included or not but this was a startup.lk error as the PV Smoothing form wasn't in the list for Fuel Cell FOM. I can push a fix if we do want to include this as an option
> 
> @mjprilliman Can you push a fix to add the PV smoothing UI form to Fuel Cell configurations for now? After my latest commit, if you choose PV smoothing a red warning message should display saying that PV smoothing is for PV Battery only. Otherwise we will need a different version of batt_dispatch_choice_ui for PV and non-PV battery configs.

@cpaulgilman This wouldn't exactly be an elegant solution but could we change the Height property on batt_dispatch_choice_ui for non PV-Battery configs so that users would never see PV Smoothing? It will not be the default for any cases so the PV Smoothing UI form would only load in the case of someone changing the value of batt_dispatch_choice_ui from the Inputs Browser, which would then be covered by your message.

cpaulgilman:
> 
> 
> > > > This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:
> > > > image
> > > > Do we want to provide the option for that configuration or remove it?
> > > 
> > > 
> > > Not sure on whether it should be included or not but this was a startup.lk error as the PV Smoothing form wasn't in the list for Fuel Cell FOM. I can push a fix if we do want to include this as an option
> > 
> > 
> > @mjprilliman Can you push a fix to add the PV smoothing UI form to Fuel Cell configurations for now? After my latest commit, if you choose PV smoothing a red warning message should display saying that PV smoothing is for PV Battery only. Otherwise we will need a different version of batt_dispatch_choice_ui for PV and non-PV battery configs.
> 
> @cpaulgilman This wouldn't exactly be an elegant solution but could we change the Height property on batt_dispatch_choice_ui for non PV-Battery configs so that users would never see PV Smoothing? It will not be the default for any cases so the PV Smoothing UI form would only load in the case of someone changing the value of batt_dispatch_choice_ui from the Inputs Browser, which would then be covered by your message.

Thinking outside of the box, eh? ;)

I'm not sure that will work across operating systems because of the difference in font size. I think you could use the property() function to conditionally set the Items property of the batt_dispatch_choice widget, but then that would not be reflected in the Index Labels for that variable, so the option would appear in Parametrics and Inputs Browser. It's also a hack that might cause confusion for someone besides us down the road trying to understand how the form works..

mjprilliman:
> > > > > This is the error I get when trying to select PVSmoothing on the Fuel Cell - Single Owner configuration:
> > > > > image
> > > > > Do we want to provide the option for that configuration or remove it?
> > > > 
> > > > 
> > > > Not sure on whether it should be included or not but this was a startup.lk error as the PV Smoothing form wasn't in the list for Fuel Cell FOM. I can push a fix if we do want to include this as an option
> > > 
> > > 
> > > @mjprilliman Can you push a fix to add the PV smoothing UI form to Fuel Cell configurations for now? After my latest commit, if you choose PV smoothing a red warning message should display saying that PV smoothing is for PV Battery only. Otherwise we will need a different version of batt_dispatch_choice_ui for PV and non-PV battery configs.
> > 
> > 
> > @cpaulgilman This wouldn't exactly be an elegant solution but could we change the Height property on batt_dispatch_choice_ui for non PV-Battery configs so that users would never see PV Smoothing? It will not be the default for any cases so the PV Smoothing UI form would only load in the case of someone changing the value of batt_dispatch_choice_ui from the Inputs Browser, which would then be covered by your message.
> 
> Thinking outside of the box, eh? ;)
> 
> I'm not sure that will work across operating systems because of the difference in font size. I think you could use the property() function to conditionally set the Items property of the batt_dispatch_choice widget, but then that would not be reflected in the Index Labels for that variable, so the option would appear in Parametrics and Inputs Browser. It's also a hack that might cause confusion for someone besides us down the road trying to understand how the form works..

@cpaulgilman Trying this with the Height property did not update the widget on the first load for whatever reason. Changing the Items property does work but I ran into the same confusion you mention. I think I will just rely on your message and leave it at that. Should I disable the PV Smoothing inputs for non PV Battery configs? Right now SAM will still run for Generic Battery or Fuel Cell and the PV Smoothing option selected, the results probably just don't make much sense

sjanzou:
I think we should disable PV Smoothing for all but detailed PV. We discussed that smoothing is only an option for PV systems. I see where smoothing may be useful for the PV Fuel Cell combo and i see some use cases for generic.

The red error message was directly from Will. Smoothing does not make sense at weather file intervals greater than 15 minutes per the EPRI meetings.

Let me know what you think and i will update the pull request.

Thanks for the detailed reviews!

mjprilliman:
@sjanzou @cpaulgilman Take a look at the latest commit and see if that is sufficient or if we need to more on non PV Battery

cpaulgilman:
> 
> 
> @sjanzou @cpaulgilman Take a look at the latest commit and see if that is sufficient or if we need to more on non PV Battery

This looks good to me. I revised the message to mention 15 minute weather data instead of 30 minutes.

sjanzou:
Looks good to me. Any other items for this pull request?

mjprilliman:
> Looks good to me. Any other items for this pull request?

Nothing from me

cpaulgilman:
> 
> 
> > Looks good to me. Any other items for this pull request?
> 
> Nothing from me

Nor from me.

sjanzou:
Generic Battery and Fuel Cell updates for FOM dispatch with PV Smoothing updated and ssc tests added and project files tested. Please recheck and let me know if you have any changes. Thanks!

deploy/runtime/startup.lk
```diff
@@ -559,9 +559,10 @@ function setup_fuel_cell_pages_fom()
 			'Battery Dispatch Front of Meter',
 			'Fuel Cell Dispatch Manual']], {'sidebar'='Dispatch', 'help'='fuelcell_dispatch'});*/
 	addpage( [[ {'name'='Battery Dispatch Automated FOM', 'caption'='Automated'} ], [ {'name'='Battery Dispatch Custom Time Series', 'caption'='Custom Time Series'} ], 
-		['Battery Dispatch Empty'],[{'name'='Battery Dispatch PV Smoothing', 'caption'='PV Smoothing'}]],
+		[{'name'='Fuel Cell Dispatch Manual', 'caption'='Manual'}],
```
mjprilliman, line 562: So @cpaulgilman and I have discussed this issue. The battery manual dispatch page for fuel cell also includes the controls for fuel cell manual dispatch. So if the user selects Manual Dispatch for fuel cell dispatch but not for battery dispatch, they will not be able to see the controls for manual fuel cell dispatch. This was why the Fuel Cell Dispatch Manual was fixed as a header page.

deploy/runtime/ui/Battery Dispatch Options FOM.txt
```diff
@@ -476,6 +480,7 @@ function warning_message()
 	{
 		msg = ''; 
 	}
+	/*
 	elseif ( value('batt_dispatch_choice_ui') == 3 ) 
 	{ 
 		if ( technology() != 'PV Battery' )
```
mjprilliman: So is the final decision to give a warning to the user that the PVSmoothing option is only intended for PV Battery configs or not?

deploy/runtime/ui/Battery Dispatch PV Smoothing.txt
```diff
@@ -1773,7 +1773,8 @@ on_load{'Battery Dispatch PV Smoothing'} = define() {
 function toggle_dispatch_pvs()
 {
 	mode = ${batt_dispatch_choice};
```
mjprilliman, line 1775: If we are hard-coding this then `mode` is not needed.

mjprilliman (commented):
A little confused now and what the final verdict is on the Fuel Cell FOM handling

test_results_win64.csv
```diff
@@ -121,7 +121,7 @@ Generic System,Sale Leaseback,756864000,32.130512,35.256783,;
 Generic System,LCOE Calculator,756864000,NA,NA,; 
 Generic System,None,756864000,NA,NA,; 
 PV Battery,Residential,8588.589844,11.205457,NA,; 
-PV Battery,Commercial,911872.062500,4.274726,NA,; 
+PV Battery,Commercial,911877.375000,4.274710,NA,;
```
mjprilliman, line 124: These are the test values that are in Patch 2. I changed the values with this commit (https://github.com/NREL/SAM/commit/6b8ebd84e106c7b5158b71d49bddbf984fbaa4c9#diff-8b237736d5563a32f81df3035be5cccb0f7b2cfd8b863b1d3d4e5956f65dfb1dR123) and must not have changed it back.

brtietz (approved):
Changes look good to me. Ok to merge, contingent on SSC issues being solved. I'm fine with resolving the manual dispatch UI form issues in another PR.

PR #666, opened 2021-07-23
Fix Defaults for PV and/or Battery 

dguittet:
Method:
 - Open a Model with a config
 - Shift+F10
 - Change to another Model
 - Restore inputs to defaults
 - Shift+F10
 - Repeat above 3 all configs
 - Use Gitkraken or some visual Git editor to check changes

Most fixes are around cost defaults, but some are related to default size of the battery system
 
Close ##665


mjprilliman:
@dguittet @cpaulgilman So I don't think the listed methods had the desired effect. I think using Change Model followed by Reset inputs to default values is keeping the defaults of the original configs, so the system capacities and costs from whatever the first config was were written to the remaining configs. Hence the failing Travis tests. I'm gonna try to go 1 config at a time to see if that has the intended effect.

mjprilliman:
Closing as this was addressed in #668

Issue #665, opened 2021-07-22
Defaults for PV & Battery Divergence


dguittet:
The defaults for FlatPlatePV Single Owner were updated recently in https://github.com/NREL/SAM/commit/992c6bdf6b2d4d9c6d69ba5e5cce947d56186a5c#diff-2c4938936666acfae618b1684ecfd2241a9bbd5fd82590de67c32a207f900c07 changing the cost defaults for many of the PV & Battery models such as the total installed cost and construction financing cost, while the system_capacity remained the same:

For example, in Flat Plate PV_All Equity Partnership Flip.txt, total_installed_cost increased from 51384528 to 5.40779e+12.

In the GUI, these values are overwritten from a callback or equation, so that total_installed_cost is around 51384528 again.

The problem is that for these files, the PySAM modules use the defaults from the .txt file (5.40779e+12) and due to these costs being different between the GUI and the .txt file (while the system config is the same), the PySAM and SAM default simulations have extremely different results. 

-------

I went and did F10 on all the PV and/or Battery configs and found a lot of the defaults to be out of date with the GUI... Is there something systematic we can do to assure the defaults file reflects the GUI as well as possible?

dguittet:
@sjanzou mentioned a script to check all the Ui forms for all the configurations will trigger all the calculations (equations and callbacks). And an option to overwrite the defaults after checking all ui forms for each configuration. I think a option for each UI variable to be excluded from being overwritten would also be helpful

sjanzou:
@dguittet, @mjprilliman is this still an issue? 

There is a "Test_UI" branch of SAM that has an additional invoke.cpp "check_configuration" that is supposed to run through a configuration's pages and open each one to test the callbacks and equations. 

I never had a chance to finish testing... would this be the desired function here? 

There is currently no option for each variable (UI or ssc) to be excluded from being overwritten.

mjprilliman:
> @dguittet, @mjprilliman is this still an issue?
> 
> There is a "Test_UI" branch of SAM that has an additional invoke.cpp "check_configuration" that is supposed to run through a configuration's pages and open each one to test the callbacks and equations.
> 
> I never had a chance to finish testing... would this be the desired function here?
> 
> There is currently no option for each variable (UI or ssc) to be excluded from being overwritten.

We pushed a PR fixing the inconsistencies for the release but there might still be a desire for that function. @dguittet could speak to that more than me.

PR #668, opened 2021-07-23
Fix calculated value defaults for PySAM

mjprilliman:
-See #665 
-PR #666 overwrote defaults for several configs causing Travis to fail
-Need to check that defaults for PV, Battery, PVwatts configs are accurately reflecting defaults for calculated values to be imported into PySAM

Methods: 

1. Open case
2. Shift F-10 (loading each page or opening/collapsing each tab was found to have no effect on defaults)
3. Repeat 1-2 with next technology-financial combination

### DIFF
--- a/deploy/runtime/defaults/MEwave_LCOE Calculator.txt
+++ b/deploy/runtime/defaults/MEwave_LCOE Calculator.txt
@@ -4082,13 +4082,11 @@ wave_resource_end_period
 wave_resource_filename
 1
 4
-37
-Humboldt_California_Wave-Resource.csv
+0
 wave_resource_filename_ts
 1
 4
-37
-Humboldt_California_Wave-Resource.csv
+0
 wave_resource_matrix
 1
 3
--- a/deploy/runtime/defaults/MEwave_None.txt
+++ b/deploy/runtime/defaults/MEwave_None.txt
@@ -528,13 +528,11 @@ wave_resource_end_period
 wave_resource_filename
 1
 4
-37
-Humboldt_California_Wave-Resource.csv
+0
 wave_resource_filename_ts
 1
 4
-37
-Humboldt_California_Wave-Resource.csv
+0
 wave_resource_matrix
 1
 3
--- a/deploy/runtime/defaults/Wind Power_All Equity Partnership Flip.txt
+++ b/deploy/runtime/defaults/Wind Power_All Equity Partnership Flip.txt
@@ -4507,8 +4507,7 @@ wind_resource_distribution
 wind_resource_filename
 1
 4
-26
-WY Southern-Flat Lands.srw
+0
 wind_resource_model_choice
 1
 1
--- a/deploy/runtime/defaults/Wind Power_Commercial.txt
+++ b/deploy/runtime/defaults/Wind Power_Commercial.txt
@@ -4331,8 +4331,7 @@ wind_resource_distribution
 wind_resource_filename
 1
 4
-26
-WY Southern-Flat Lands.srw
+0
 wind_resource_model_choice
 1
 1
--- a/deploy/runtime/defaults/Wind Power_LCOE Calculator.txt
+++ b/deploy/runtime/defaults/Wind Power_LCOE Calculator.txt
@@ -1128,8 +1128,7 @@ wind_resource_distribution
 wind_resource_filename
 1
 4
-26
-WY Southern-Flat Lands.srw
+0
 wind_resource_model_choice
 1
 1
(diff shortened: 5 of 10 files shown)


deploy/runtime/defaults/Flat Plate PV_Merchant Plant.txt
```diff
@@ -1100332,19 +1100332,19 @@ total_direct_cost
...
 1
-5.15027e+12
+42746896
 total_indirect_cost
 1
 1
 1
 1
-6500290
+6.50029e+06
```
mjprilliman, line 1100341: total_indirect_cost no longer includes sales tax. Total installed is now the sum of total_direct_cost, sales_tax_total, and total_indirect_cost

dguittet (changes requested):
Thanks Matt!

I noticed changes to `step`, `step_spe`, and `ui_step_minutes`. Will these cause issues?

Also, please remove change in PVWatts_None.txt:

```
solar_data_file_name_load
1
4
49
phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv
```

mjprilliman (commented):
> Thanks Matt!
> 
> I noticed changes to `step`, `step_spe`, and `ui_step_minutes`. Will these cause issues?
> 
> Also, please remove change in PVWatts_None.txt:
> 
> ```
> solar_data_file_name_load
> 1
> 4
> 49
> phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv
> ```

So `step` is an output from Wfreader that is basically the number of seconds in a time step from the file. It must have changed that default based on the default weather file. Should that default remain zero? `ui_step_minutes` just converts the seconds to minutes to display in the UI and is used in a callback check for whether or not the Battery Reopt optimization can proceed. 

`step_spe` is an unused copy of step and should probably be taken out of the defaults. 

I will take those lines out of PVWatts_None.txt

deploy/runtime/defaults/PVWatts_None.txt
```diff
@@ -322,7 +322,8 @@ phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv
 solar_data_file_name_load
 1
 4
-0
+49
+phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv
```
mjprilliman, line 326: @dguittet @cpaulgilman should solar_data_file_name and solar_data_file_name_load not have a default string? The changes are not showing on most files because the default weather file is already loaded into the defaults file.

dguittet: Ah, I guess that's fine if there's a default file_name in the UI, especially if that's already how it is. What I need to avoid is SSC file names variables being stored, so `solar_resource_file` must be empty.

deploy/runtime/defaults/Flat Plate PV_Merchant Plant.txt
dguittet: gotcha

deploy/runtime/defaults/PVWatts_None.txt
mjprilliman: I'm assuming that applies to every config with solar_resource_file? It seems that even the CSP configs have the default string. I just pushed that change for PV configs

cpaulgilman (approved):
I ran my test script to run all default configs, and seemed ok. I did get one message for MWwave/LCOE Calculator "variable 'file_name' not found in collection" but could not replicate it, so should be OK for now.

deploy/runtime/defaults/PVWatts_None.txt
dguittet: Ah, well, it's unavoidable then :D I'm okay with leaving them for the CSP and others. Thanks for fixing those!

dguittet (approved):

deploy/runtime/defaults/PVWatts_None.txt
mjprilliman: I can push a commit blanking out solar_resource_file for all configs that have it if you'd like. 88 additional default files from the ones already in this PR but its all the same change.

dguittet: Thanks Matt but I think it's not worth the time, unless you have already an automated way to do it?

dguittet: If not then don't worry about it! I've blanked it out before a year or so ago and it didn't last. It doesn't really create a bug, it's just I prefer users, when they try to run a default case in PySAM, to get an error about "you need to add a solar resource file" rather than "C:\dguittet\SAM\weatherfile\x.csv" doesn't exist.

mjprilliman: I used the Defaults Manager on every single configuration. It should have modified the default in configs where solar_resource_file did exist and not do anything to those that do not have it. You are right though it could easily get overwritten again. 

image

PR #669, opened 2021-07-27
Cambium api additions

StephenPaul27:
I just merged develop, fixed conflicts, and ran the test script, the results came out clean and everything seems to be working fine so let me know if there's anything you spot that needs fixing/updating.

LK/WEX/SSC branch names are all also cambium_api_additions, I'll create separate pull requests for those I guess


cpaulgilman:
1. Please remove colons from UI variable labels to be consistent with rest of UI:

image

2. The on_load() function for Cambium API Data and Cambium Tab UI forms appear to be similar. Are both required, or can that code be consolidated in one place to make it easier to maintain?

3. The urls for the Cambium API calls are hard-coded in the UI callback LK script instead of being stored in the runtime/webapis.conf file and read with the `webapi()` function. See Solar Resource and Marine Wave Resource Time Series forms for examples of the `webapi()` function.

4. Is the `project_uuid` variable to store an NREL Developer API key? If so, I think it should use the SAM API key with code similar to this code snippet from the Solar Resource UI callback. For official release versions of SAM, the `curl()` function should convert <SAMAPIKEY> to the SAM's NREL Developer API key (this only works for versions of SAM built from SAM-private):

```
base_url = dataset[n].url;
base_url += '?api_key=<SAMAPIKEY>';
base_url += '&full_name=sam_user';
base_url += '&email=<USEREMAIL>';
```

5. Is the Financial Merchant Plant form still needed? If not please remove it by deleting the .txt file in the ui folder.

StephenPaul27:
1, 3, & 5: Thanks for pointing these out, I'll fix them today.
2: The on_load was the only function I found necessary to duplicate when creating the "dummy" UI forms for the tabs because of the difference in the form name triggering the load.  I can try to either consolidate or reduce what is there currently.
4: The project_uuid is specific to Cambium and just differentiates which data set we're calling to.  
cambium_screenshot
So the project_uuid in the UI specifies that we're getting data from that first set of scenarios "Cambium Scenarios (Hourly and annual)"
Let me know if this should be relocated somewhere.

cpaulgilman:
Thanks, Stephen. The project_uuid is fine where it is -- I just wanted to make sure we are not storing API keys in the UI callback code.

deploy/runtime/webapis.conf
```diff
@@ -2,6 +2,8 @@
 ' Modifying this file may cause SAM's download features to stop working.
 
 android_build=https://sam.nrel.gov/sites/default/files/content/mobile/android/readme.html
+cambium_meta=https://cambium.nrel.gov/api/project/detail/?uuid=579698fe-5a38-4d7c-8611-d0c5969b2e54
```
sjanzou, line 5: Can we place the uuid in private.h in the SAM-private repo? Was this uuid assigned by Cambium and does it have a monthly bandwidth limit?

deploy/runtime/ui/.txt
```diff
@@ -0,0 +1,9 @@
+
+500
```
sjanzou, line 2: Can you remove this .txt file before merging?

sjanzou (changes requested):
Please remove the .txt file per comments and move the uuid to private.h in SAM-private as used for the <SAMAPIKEY> so that the SAM Cambium key is "hidden" from webapis.conf. What will the  final defaults used be - hourly, single value and then populated by Cambium? Great Work!

deploy/runtime/webapis.conf
StephenPaul27: The uuid just differentiates which Cambium Data Viewer project that data is pulled from.  I don't think it's necessarily the same as an API key, and I don't know if they have any monthly bandwidth limits on their projects.  I can ask Robert about that.  I can place it in the SAM-private repo if you think that would be better or remove confusion later though.

StephenPaul27: I confirmed that there is no monthly bandwidth limit currently on the Cambium projects.

PR #677, opened 2021-08-08
Fixed clipped system power battery charging option for Generic Battery systems

### LINKED ISSUE #646: Generic system shows "battery can charged from clipped power" which doesn't make sense for AC connected systems
Generic systems must be AC-connected, so charging from clipped power is not an option. However, the box shows up as checked by default:

image

Box should be unchecked in generic systems. Exists in both 2020.11.29, and on develop.

mjprilliman:
-Changed State default from 1 to 0 as no clipped power goes to battery for AC connected Generic Battery systems
-Added State check in callbacks 
-Closes #646


brtietz (approved):
Looks good! Thanks for doing this fix!

PR #679, opened 2021-08-09
Removed weather file ambient temp option for battery room temperature

### LINKED ISSUE #632: Remove battery option to use weather file temperature
### Problem

For the battery environmental temperature option, the "Use weather file ambient temperature" option is unrealistic because batteries are typically installed in a temperature-controlled container or room, so the battery is not exposed to ambient temperatures. Having that option is potentially misleading for new users.

If someone really wanted to use the ambient temperature as the battery environment temperature, they could use the "Enter time series temperature" option.

### Resolution

Remove "User weather file ambient temperature" option and add code in update script to handle importing old files with that option enabled: Warning message if weather file ambient temperature option is selected, and change value of  `batt_room_temperature_vector` as needed for other options.

This will potentially break LK scripts and other code created in older versions because `batt_room_temperature_vector = 0` is the weather file temperature option.

mjprilliman:
-Closes #632 
-Add change in battery room temperature options to Upgrade script
-Remove option for weather file ambient temperatures to be used as battery room temperature inputs on Battery Thermal page

### DIFF
--- a/deploy/runtime/defaults/Generic Battery_All Equity Partnership Flip.txt
+++ b/deploy/runtime/defaults/Generic Battery_All Equity Partnership Flip.txt
@@ -925,7 +925,7 @@ batt_thermal_choice
 1
 1
 1
-1
+0
 batt_time_capacity
 1
 1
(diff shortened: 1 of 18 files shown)


sjanzou (approved):
Looks good, thanks for updating the version.lk script, too!

brtietz:
Not sure why git is saying this is unmerged. I merged this into develop yesterday in 15a1f567ef8214dff974e3d21835cd424dda2c4a

Issue #673, opened 2021-08-02
Code generator broken on develop


brtietz:
### Describe the bug

On develop (the 2021.7.27 beta) the code generator is broken with the error "input ordering failed" - however, it appears the expected files are generated anyway.

### To Reproduce

1. Build the latest develop
2. Open any case
3. Export using shift-F5
4. Choose any language
5. See "input ordering failed error"

### Expected behavior

The code generator should open the folder where the files were generated rather than throwing an error.

### Screenshots

![image](https://user-images.githubusercontent.com/5530592/127899483-757f3384-b1ab-4cd1-a36f-eca38f3cccc2.png)

### Operating System and Version

 - OS: Windows
 - Version: 2021.7.27 (and maybe earlier on develop)

sjanzou:
Closed with pull requests #708 and #707

PR #684, opened 2021-08-17
Fixes SAM issue #673

sjanzou:


brtietz (changes requested):
It looks like this code leaves a mismatch between the compute module name and the defaults exported. This may be fine for many forms of export, but it breaks PySAM JSON. For example, exporting a PV SingleOwner case will put the SingleOwner outputs in utilityrate5, and not export single owner. Similarly, a residential case will not export cashloan:

image

sjanzou:
Fixed with pull request #708  https://github.com/NREL/SAM/pull/708 and merged into critical-load-ui which was then merged into develop with pull request #707 https://github.com/NREL/SAM/pull/707

PR #686, opened 2021-08-17
Fix Geothermal crashing issue

### LINKED ISSUE #639: Reopening SAM file containing Geothermal case with cost model option crashes SAM
### Describe the bug

When saving a SAM file with a Geothermal case, if the 'Automatically estimate the plant cost per kW' box is checked on the System Costs page, reopening the saved file crashes SAM. The 'ui_calculations_only' mandatory input to the geothermal compute module seems to be the issue. 

### To Reproduce

1. Create a Geothermal case in a new SAM file
2. Check the 'Automatically estimate the plant cost per kW' checkbox on the System Costs page under Plant Capital Cost
3. Save the file as a .sam file
4. Close the file
5. Attempt to reopen the file

### Expected behavior

SAM should not crash when reopening these types of SAM files

mjprilliman:
-Closes #639 
-Option to calculate $/kW value for Geothermal costs no longer crashes SAM when reopening saved file


cpaulgilman (changes requested):
This fixes the bug, but I'm seeing this message about defaults:

No default value found for 'baseline_cost' in external file (Geothermal Power/Single Owner), using internal default

Thanks,
Paul.

Issue #634, opened 2021-06-09
Add UI function that auto-populates cmod inputs


tyneises:
It can be helpful to call the technology compute module from a UI page script to calculate design parameters from the same SSC code that the annual simulation will use. Currently this is possible by defining each compute module input with its matching UI variable (e.g., ssc_var(obj, "tshours", ${tshours});). However, this is tedious and causes opaque ui crashes when a new input is added to the compute module but not assigned in the UI script.

Proposed modification:
A new function that
1) Automatically populates the SSC object by assigning ui variables to compute module inputs with the same name
2) Passes a flag to the compute module indicating the type of simulation. I'm currently using an optional compute module input 'sim_type' and setting it to 2 in the ui script when I want to call the compute module for design-point calcs and not an annual simulation. This variable can't exist in the UI, because it will affect normal SAM simulations
3) Runs the specified compute module and returns compute module outputs the same as the existing 'ssc_exec' function, (e.g., ${total_installed_cost} = ssc_var(obj, "total_installed_cost");)

The function might look something like:
ssc_auto_exec(obj_solved, 'cmod_name', sim_type)

tyneises:
You can see how I'm calling the technology compute module in UI scripts the Tedious Way by:
- checking out the etes_dispatch SSC branch: https://github.com/NREL/ssc/tree/etes_dispatch
- checking out the etes_dispatch SAM branch: https://github.com/NREL/SAM/tree/etes_dispatch
- use the UI editor to review any of the "ETES_..." ui pages
- run the ETES technology model, which is available under the CSP technology drop-down

sjanzou:
@tyneises , that is tedious! I will try to have a pull request for you soon!

tyneises:
@sjanzou I modified the UI pages to call the cmod once in the System Design page equations. Then, the System Design page stores all relevant cmod outputs. The remaining UI pages use a distinct variable name for its UI object, which it sets to the corresponding System Design output via equation{'name'} = define() {......} https://github.com/NREL/SAM/commit/c95bf3e3de51ffcc29dd3f3be7b9cbfff7510161

sjanzou:
Closed with pull request #687 into etes_dispatch which was later merged into develop

PR #687, opened 2021-08-19
ssc_auto_exec created to address #634 only works in callbacks and NOT MIMOs

sjanzou:
The ssc_auto_exec works in callback scripts but not MIMOs. For example, see ETES System Design page. The issue is that equations are evaluated before the case is created and the invoke functions need to have the case to iterate over the variables. That is why the tedious MIMO method works and updates whenever any of the specified inputs are changed. Still investigating alternate solutions.


tyneises (changes requested):
Thanks, I see the problem. The callback alternative also has this problem after the page loads - it doesn't know which variables should trigger re-evaluation of the callback. Is it possible to build a function that returns True if any variable in the cmod changes in the UI? E.g. on_change_any_cmod_input()? Or equivalently a $MIMOcmod$ that checks all cmod inputs?

sjanzou:
Search through lk parser and equation evaluator and setting of inputs and output and dependent varibles and find no lk function that will properly setup and parse equations in Case->SetConfiguration.

Alternative is to parse the equation script at input page load and expand as 
image
For proper triggering when a MIMO input changes.

Cannot handle the if financing exception above but all other values can be set and the data opject can be created and returned and all will be treated as inputs.

Change would have to occur either at 
image
or at
image

sjanzou:
@tyneises , working now in equations in SAM_634 branches of lk, ssc, and SAM. SSC variables not in UI need to have UIHINT set to SIMULATION_PARAMETER
image

tyneises (approved):
This is really great! Is there a way to determine whether the input is a SIMULATION_PARAMETER based on the value on the financial model type, like how "required_if" works? It would be convenient to eliminate the if(financial_model) logic in the MIMO on the System Design page.

PR #690, opened 2021-08-23
Batt_dispatch_choice update to switch choice to Automated for FOM

mjprilliman:
-Switching batt_dispatch_choice_ui radio buttons from another option to Automated does not update batt_dispatch_choice as it should. 
-Battery model is still running prior selection for the dispatch option
-This fix should fix that behavior


brtietz (approved):
Good catch!

mjprilliman:
> It looks like https://github.com/NREL/SAM/pull/712/files has the same fix as this branch - should this be closed once that one is merged?

I would say yes. I did not realize I never merged this in but yes I will close

PR #691, opened 2021-08-24
Batt life model GUI

dguittet:
- Add buttons for different types of battery life model; greying out the calendar and cycling widgets when they are not in use
image

- If chem = 0 for lead acid, then this new widget is disabled, since only calendar and cycling model is allowed

- Adapt chemistry-specific default values to use this widget.

- Update technology-financial defaults

- Update the test results for windows and osx

Goes with https://github.com/NREL/ssc/pull/608


deploy/runtime/defaults/PV Battery_Third Party.txt
```diff
@@ -887,6 +887,12 @@ batt_inverter_efficiency_cutoff
 1
 1
 90
+batt_life_model
+1
+1
+1
+1
+1
```
mjprilliman, line 895: Just checking is the default lifetime model supposed to be 1 for "NMC Gr Life Model" for all configs except Fuel Cell?

deploy/runtime/ui/Battery Lifetime.txt
```diff
@@ -971,14 +1026,36 @@ on_load{'Battery Lifetime'} = define() {
...
-
+	batt_chem = value('batt_chem');
+	if (batt_chem < 1) {
+		value('batt_life_model', 0);
+		enable('batt_life_model', false);
+	}
+	else {
+		on_change{'batt_life_model'}();
+		enable('batt_life_model', true);
+	}
```
mjprilliman, line 1037: This enables/disables on the page load, but you need to add this check to `on_change{'batt_type'}` or `on_change{'batt_chem'}` so the page will update without having to reload the Battery System page. Right now changing batt_chem doesn't disable batt_life_model until you go to a different tab then return.

test_results_win64.csv
```diff
@@ -120,26 +120,26 @@ Generic System,All Equity Partnership Flip,756864000,32.812180,36.422478,;
...
+PV Battery,Residential,8588.509766,11.206270,NA,; 
+PV Battery,Commercial,911876.750000,4.274711,NA,; 
+PV Battery,Third Party,34803.347656,NA,NA,; 
+PV Battery,Host Developer,911876.750000,13.250633,10.808079,; 
+PV Battery,Single Owner,92853440,7.794683,9.771930,; 
+PV Battery,Merchant Plant,92853536,6.917604,NA,; 
+PV Battery,Leveraged Partnership Flip,92853536,7.926891,9.771934,; 
+PV Battery,All Equity Partnership Flip,92853536,9.312499,10.857704,; 
+PV Battery,Sale Leaseback,92853536,9.510679,9.771934,; 
+PVWatts Battery,Residential,8022.691406,11.707511,NA,;
```
mjprilliman, line 132: Are these value changes expected when using the NMC Lifetime model instead of the Calendar-Cycle definitions?

mjprilliman (approved):
Just need to fix the enabling/disabling of batt_life_model widget to happen on batt_chem change rather than just on page load. Otherwise looks good!

deploy/runtime/ui/Battery Lifetime.txt
dguittet: Since `batt_chem`  isn't on the Lifetime page as it's on the Battery Model.txt, it doesn't seem to trigger off `batt_chem` being changed either. Is it supposed to work?

dguittet: Nevermind, adding code Battery Model.txt seems to disable batt_life_model correctly

mjprilliman: Yep that works like I would expect it to now.

deploy/runtime/ui/Battery Lifetime.txt
```diff
@@ -686,6 +686,47 @@ Cycle Degradation
...
+5
+0
+Selection
+3
+0
+Items
+6
+3
+Calendar - Cycle Table
+NMC/Gr Life Model
```
brtietz, line 719: If the NMC model is going to be the default, should there be a warning when the user changes away from that chemistry/voltage model?

mjprilliman: Maybe some kind of messaging in the upgrade script as well?

dguittet: When they change to any chemistry besides NMC and LMO/LTO, it goes back to Calendar - Cycling. Voltage model changes don't do anything

dguittet:
@cpaulgilman Thanks Paul! Let's discuss next week. I think we can rename the calendar equation to something generic-- maybe Calendar rate equation? And yes, the new models do calendar and cycling both themselves.

PR #692, opened 2021-08-24
More flexible powerflows

brtietz:
Restrictions in SAM's powerflow code required that AC connected, behind the meter batteries only charge when the renewable system's power exceeded load, and only discharge when the load exceeded the renewable system's power. DC connected batteries were subject to the latter restriction as well.

In order to accurately process optimal dispatch from REopt, these powerflow constraints need to be lifted, as some timesteps will violate these constraints. This increases the power that SAM deliveres from the battery to the load while running a REopt dispatch case from 93% of REopt's prediction to 99.94% - and should help with other custom dispatch cases as well.

Pairs with SSC PR: https://github.com/NREL/ssc/pull/609


cpaulgilman:
@brtietz For BTM batteries, should the "Charge from system only when system output exceeds load" check box be enabled only if "Battery can charge from system" is checked?

Also, I plan to add conditional text similar to the FOM dispatch form to explain options that depend on whether the battery is AC- or DC-connected.

brtietz:
> 
> 
> @brtietz For BTM batteries, should the "Charge from system only when system output exceeds load" check box be enabled only if "Battery can charge from system" is checked?
> 
> Also, I plan to add conditional text similar to the FOM dispatch form to explain options that depend on whether the battery is AC- or DC-connected.

Great point! I added enabling/disabling code based on the charge variable in the latest push.

Thanks for offering to add the conditional text!

PR #696, opened 2021-08-30
Updated test results for LCOE equation changes

### LINKED ISSUE #532: Negative LPPA and LCOE for storage projects
SAM 2020.11.29 uses net energy to grid to calculate the LCOE and LPPA. For stand alone batteries which charge from the grid, this can lead to negative LCOE and LPPA values. Determine what kWh values should be used for these calculations in cases where the net energy is negative.

mjprilliman:
-Updated LCOE, LPPA test results for changing LCOE denominator from energy net to energy sales
-Closes #532


test_results_win64.csv
```diff
@@ -17,69 +17,69 @@ PVWatts,Host Developer,943704.187500,14.690550,15.743791,;
 PVWatts,Single Owner,87966056,7.918575,8.357705,; 
 PVWatts,Merchant Plant,87966056,7.181620,NA,; 
 PVWatts,Leveraged Partnership Flip,87966056,8.091964,8.778845,; 
-PVWatts,All Equity Partnership Flip,87966056,9.820335,11.885724,; 
+PVWatts,All Equity Partnership Flip,87966056,0,0,;
```
brtietz, line 20: LCOE and LPPA should not be zero. This suggests a bug somewhere.

brtietz (changes requested):
It looks like All Equity Partnership has a bug - LCOE and LPPA should not be zero for that configuration.

test_results_win64.csv
mjprilliman: Chunk of code calculating cf_energy_sales and cf_energy_purchases for each year was not brought into ssc/cmod_equpartflip.cpp. The LCOE and LPPA are no longer zero for this financial model.

cpaulgilman (approved):

cpaulgilman:
Is the intent of this pull request to remove the option choose PPA price or retail rates for FOM battery configurations on the Electricity Purchases page? I think we want to keep that option, but may have missed something in one of the LCOE/LCOS discussions.

image

mjprilliman:
> Is the intent of this pull request to remove the option choose PPA price or retail rates for FOM battery configurations on the Electricity Purchases page? I think we want to keep that option, but may have missed something in one of the LCOE/LCOS discussions.
> 
> image

I did think the intention was to remove this option as I thought we had decided to always treat these FOM electricity purchases as operating costs rather than PPA revenue deductions. Thus I started the process of introducing a button to convert the PPA rate into a retail rate if the user wanted to base the retail rate used to calculate operating costs of parasitic loads off of the specified PPA price. I suppose if we left in the 'Use PPA price' option we could calculate the retail rate from the PPA price behind the scenes. Would that be preferred?

cpaulgilman:
> 
> 
> > Is the intent of this pull request to remove the option choose PPA price or retail rates for FOM battery configurations on the Electricity Purchases page? I think we want to keep that option, but may have missed something in one of the LCOE/LCOS discussions.
> > image
> 
> I did think the intention was to remove this option as I thought we had decided to always treat these FOM electricity purchases as operating costs rather than PPA revenue deductions. Thus I started the process of introducing a button to convert the PPA rate into a retail rate if the user wanted to base the retail rate used to calculate operating costs of parasitic loads off of the specified PPA price. I suppose if we left in the 'Use PPA price' option we could calculate the retail rate from the PPA price behind the scenes. Would that be preferred?

Sorry for not tracking what decisions have been made on this -- I think there is a legitimate modeling scenario for PPA projects without storage where the cost of small nighttime loads such as for inverter standby is modeled as negative PPA revenue, but no need to rehash the decision to change that if we already made it.

PR #704, opened 2021-09-23
Me macro updates

mjprilliman:
-ME Tidal report macro
-ME Tidal compares cases macro
-Similar to ME Wave report and compares cases macros


cpaulgilman:
I'm having issues with the macro intermittently not working: The browse() function opens the web browser, but the browser displays a File Not Found error. Are you seeing that behavior?

mjprilliman:
@cpaulgilman I did come across that but I think I only ran each macro once. I will investigate.

cpaulgilman (approved):
Seems to be working fine now.

PR #706, opened 2021-09-27
Combine cases

Matthew-Boyd:
This is the completed Combined Cases feature as migrated from the macro. An example usage would be to:

1. Open new project
2. Add a default PVWatts - Single Owner case
3. Add a default Parabolic Trough - Empirical - Single Owner case
4. Add a Generic System - Single Owner case
5. Select: Generic System -> Power Plant -> Generate production profiles from open cases -> Select cases...

This has been tested to combine all technology and financial models, and works with both the Generic System and the Generic System-Battery models.

Help documentation has been sent to Paul.

@sjanzou  There is one problem when using this with the Power Tower Molten Salt model. The progress bar showing the simulation of the flux maps fails to close on its own and requires the user to press the Close button. I think this has to do with the last call in Simulation::Invoke(), but I wasn't able to rectify it. Do you have any thoughts here? Thanks for reviewing.


sjanzou:
Increased height of degradation input
image

Power Tower - simulating flux maps per comment
image

Only remaining power tower message (which should probably remain):
image

These changes are in pull request https://github.com/Matthew-Boyd/SAM/pull/4

Please update this pull request after you have a chance to review my pull request - then go ahead and merge - looks great!

sjanzou (changes requested):
Please review https://github.com/Matthew-Boyd/SAM/pull/4 and then merge after you take whichever changes seem appropriate. 

Great work!

Matthew-Boyd:
@sjanzou  Thank you for the review. I think you may not have seen the latest code, however. For example, I changed the wxSpinCtrlDouble that was too short to a AFSchedNumeric . Maybe try pulling again? Regardless, the new AFSchedNumeric control appears to be the same height as the other same controls in SAM.

Regarding the fix for the Power Tower, adding 'true' to bcsim.Invoke(true) suppresses all of the progress bars, which can make it appear that SAM hangs. I think making the user press 'OK' just for the power tower model is a better tradeoff? So I'm thinking not add the 'true' argument.

sjanzou:
> @sjanzou Thank you for the review. I think you may not have seen the latest code, however. For example, I changed the wxSpinCtrlDouble that was too short to a AFSchedNumeric . Maybe try pulling again? Regardless, the new AFSchedNumeric control appears to be the same height as the other same controls in SAM.
> 
> Regarding the fix for the Power Tower, adding 'true' to bcsim.Invoke(true) suppresses all of the progress bars, which can make it appear that SAM hangs. I think making the user press 'OK' just for the power tower model is a better tradeoff? So I'm thinking not add the 'true' argument.

@Matthew-Boyd , sounds good... I wanted to provide you with the options in SAM and adding the wxBusyCursor to, at least, show that SAM was potentially not hanging. The pull request is good and approved by me ;-)

Matthew-Boyd:
Thanks @sjanzou !

PR #708, opened 2021-09-28
Fix broken code generation to test broken peak shaving

sjanzou:


brtietz:
When testing with the default pv-battery commercial case and trying to export PySAM JSON, I get this:

image

However, this was against the connect_outage_vars_to_powerflow branch on SSC. Is the Fix_issues branch with the change to batt_annual_charge_energy required for that to work?

brtietz (approved):
Latest version looks great, thanks!

PR #712, opened 2021-10-12
Updated battery dispatch options for custom time series dispatch

### LINKED ISSUE #711: Charge control options missing from custom dispatch in develop
### Describe the bug

Dispatch control from 2020.11.29:

image

same screen for develop:

image

The user needs to have the ability to set whether the battery can charge from grid and/or system for custom dispatch.

There might be issues with the upgrade script here too - when I loaded a grid charging only case from 2020.11.29, both boxes were unchecked.

mjprilliman:
-Fixed callback check for battery charging options for custom time series
-Added missing batt_dispatch_choice update for automated dispatch option
-Closes #711


brtietz (approved):
This solved the problem, thanks!

PR #714, opened 2021-10-14
UI updates for battery lifetime

cpaulgilman:
Changes include:

* Move battery lifetime and replacement inputs to new Battery Life page.

* Fix updating of battery life defaults when user changes battery type and resets to default.

* Reorganize battery life inputs and add conditional UI message to make the choices easier to understand.

* Add message to upgrade script explaining inputs have moved.

I will update Help in Develop.

SAM file for testing upgrade script
[battery-life-2020-11-29-upgrade-script.zip](https://github.com/NREL/SAM/files/7349673/battery-life-2020-11-29-upgrade-script.zip)
:

### DIFF
--- a/deploy/runtime/startup.lk
+++ b/deploy/runtime/startup.lk
@@ -1943,14 +1943,15 @@ function setup_pv_battery_pages(is_btm)
 		setup_pages_battery_cell_and_system(true); // true includes UI form for REopt
 		setup_pages_battery_life();
 		setup_pages_battery_dispatch_btm();
+		setup_grid_page_with_outage();
 	}
 	else
 	{
 		setup_pages_battery_cell_and_system(false); // no REopt for FOM
 		setup_pages_battery_life();
 		setup_pages_battery_dispatch_fom();
+		setup_grid_page();
 	}
-	setup_grid_page_with_outage();
 }
 
 setconfig( 'PV Battery', 'Residential' );
@@ -2041,14 +2042,15 @@ function setup_generic_battery_pages(is_btm) {
 		setup_pages_battery_cell_and_system(false); // false excludes UI form for REopt, only available for PV battery
 		setup_pages_battery_life();
 		setup_pages_battery_dispatch_btm();
+		setup_grid_page_with_outage();
 	}
 	else
 	{
 		setup_pages_battery_cell_and_system(false); // no REopt for FOM
 		setup_pages_battery_life();
 		setup_pages_battery_dispatch_fom();
+		setup_grid_page();
 	}
-	setup_grid_page_with_outage();
 }
 
 setconfig( 'Generic Battery', 'Residential' );


deploy/runtime/versions.lk
```diff
@@ -63,6 +63,7 @@ version_upgrade{'2021.7.27'} = define() {
 	old_temp_celsius = oldvalue('batt_room_temperature_celsius');
 	new_temp_vector = [];
 	if (is_batt_tech) {
+		message('The battery replacement and lifetime inputs are now on the new Battery Life input page. Battery life inputs change to the appropriate defaults when you change the battery type on the Battery Cell and System page and accept the default values.');
```
brtietz, line 66: I think the versions.lk script needs to include the variable "batt_life_model" as well. My preference would be to keep upgraded files with the same settings they had prior to the upgrade (cycle-life degradation for everything, I think) and have users to actively choose to use one of the new models.

deploy/runtime/ui/Battery Life Options.txt
```diff
@@ -0,0 +1,307 @@
...
+		else // other Li-ion types
+		{
+			best_life = 'empirical cycle and calendar degradation'; 
+			if ( ${ui_batt_life_model} != 0 || ${batt_calendar_choice} != 1 ) { clr = 'Red'; }
+		}
+	}
+	else // lead acid and flow
+	{
+		best_life = 'cycle degradation with no calendar degradation';
+		if ( ${ui_batt_life_model} != 0 || ${batt_calendar_choice} != 0 ) { clr = 'Red'; }
```
brtietz, line 301: I can see why we wouldn't want lead acid and flow batteries using the empirical model but is there a reason we are disabling custom?

image

A search on this suggested a lead acid battery stored at less than 100% state of charge would degrade via sulfation - though a table of days and capacity wouldn't fully model that mechanism.

brtietz (changes requested):
Really cool how this does updates across pages! A few notes and questions from me prior to merging.

In addition to the ones inline: it looks like the new battery life page needs to be added to the fuel cell model.

deploy/runtime/versions.lk
cpaulgilman: Good catch on this and the fuel cell behavior! Commit 29b48d25b should fix setting batt_life_model in the versions.lk script.

Commits f471ff0c1 and f3e60f26c should fix the fuel cell model using the old Battery Lifetime form because the Enable Battery choice requires that all battery inputs be on one page.

deploy/runtime/ui/Battery Life Options.txt
cpaulgilman: This is fixed in f3e60f26c: SAM yesno message when the user chooses the empirical option for a lead-acid or flow battery.

brtietz:
Thanks for doing those fixes! Things I noticed as of 663014d

Doing the version upgrade leaves things in a weird state on the batt life page:

image

I expect if I run this it would use the correct values, but I have to click off of the cycle-life model and back on to it to get the curves to appear.

It looks like there might be a defaults conflict on the detailed PV behind the meter models:

image

Finally, this message pops up when I select "custom" - I wonder if the message saying "empirical" is confusing? 

image

If it's too much work to differentiate between empirical and custom for the popup, maybe say "The calendar degradation model..."

brtietz:
Follow up the the grid outage screenshot - the outage input has disappeared from the grid page for those models on this branch:

image

The problem also exists in generic battery. PVWatts-Battery has the expected behavior.

PR #720, opened 2021-10-28
Fix battery sizing behavior when switching between kWh and h definitions

### LINKED ISSUE #716: Parametrics interface does not properly sizing batteries for duration
### Describe the bug

When attempting to do parametric run on battery duration in hours, the battery strings and cells are not updated appropriately.

### To Reproduce

1. Set up a detailed pv-battery model
2. Select "hours" for the duration input on the cell and system page
3. Set up a parametrics run varying the number of hours with battery back installed capacity as an output
4. See that installed capacity is the same regardless of the input hrs

### Expected behavior

The parameteric runs should vary based on the input hours

### Screenshots

image

Occurs in 2020.11.29 r2 and on develop

mjprilliman:
-Make batt_bank_size a calculated variable based on either kWh UI entry or batt_bank_duration * batt_bank_power
-Update file defaults with new batt_bank_size_ui variable for displaying kWh definition in UI
-Closes #716


brtietz (approved):
Looks great, thank you!

cpaulgilman (approved):
Looks good to me.

PR #728, opened 2021-11-03
Apply ratchets to per kw charges

brtietz:
Pairs with SSC PR https://github.com/NREL/ssc/pull/637

- Removes references to energy charges from billing demand forms, field names, and equation names
- Adds variable ur_dc_billing_demand_periods - the defaults have billing demand on for all TOU periods


brtietz:
Test results match develop. We're out of Travis credits, so I'm not sure when I'll be able to get a passing test result. I can post the results of test_script_ow.lk after develop is updated and merged in if desired.

cpaulgilman (changes requested):
Looks good. 

I would change the default value of `chk_update_peaks` to 1 so that the Year 0 peaks match the load data if a new user changes load data before going to the Electricity Rates page.

PR #730, opened 2021-11-04
New community solar UI and defaults

cpaulgilman:
Goes with https://github.com/NREL/ssc/pull/638


sjanzou (approved):
Looks good, I can re-review after you have a chance to update the defaults now that the develop branch default manager is working.

janinefreeman (changes requested):
Super exciting! Looks good overall! A few minor comments before merging: 

- The error message on the community solar page when the system is oversubscribed still refers to "subsidized" subscribers
- Should the notices about how oversubscribed projects were reallocated be a warning instead? You may have already discussed that so I don't think it's required if you decided against it.
- The x axis labels look funky on the "subscriber share of system" graph on the summary page

cpaulgilman:
Thanks @janinefreeman! Items 1 and 3 are fixed in SAM Commit 78be4e5e3, and Item 2 is fixed in SSC Commit 66a156777.

PR #733, opened 2021-11-08
adding dimensions for Silfab SIL-330BL

burundiocibu:
While using pvlib to model a system I found the SIL-330BL wouldn't load. It appears because of the missing panel dimensions. I've added the dimensions of the active area of the panel; i.e. excluding the frame.


cwhanse:
@burundiocibu I don't think the missing values prevent the module data from loading. I suspect its a pvlib usage issue. Can you post some code in a new [Discussion ](https://github.com/pvlib/pvlib-python/discussions) at the pvlib repo?

cpaulgilman:
The length and width parameters in SAM's module library are optional -- that data is not provided for all modules in the Excel file for the the CEC equipment lists: https://solarequipment.energy.ca.gov/Home/PVModuleList.

We build a new SAM module library from the CEC Excel file at the time of each release and overwrite this file with a new version with data as provided by the CEC.

PR #734, opened 2021-11-08
SAM issue 550

sjanzou:


cpaulgilman (approved):
I added debt fraction to the Metrics table, so either minimum DSCR or debt fraction is displayed.

PR #737, opened 2021-11-09
Community Solar Starup Fix

sjanzou:


sjanzou:
This is a result of Github "This branch has no conflicts with the base branch" merging. The community solar configuration still had the Financial Return on Equity form in the startup.lk configuration and that form had been removed. So when starting:
image

the configuration simulated without issue. The defaul configurations were run again and several were updated because of removed roe_input variable that was merged into develop to fix issue 549 (after develop was merged into that branch of SAM).

brtietz (approved):
Looks good to me.

I'd say reporting a clean merge that actually has some problems is an issue with git merge generally, and not GitHub specifically. Almost everything in the SAM repository is better off being merged locally, so there's a chance to test for this sort of thing prior to it hitting develop.

PR #740, opened 2021-11-09
Tidal devices updates

mjprilliman:
-Added second default option for tidal devices
-Can import power curve from csv or clipboard
-No writing to library currently

### DIFF
--- a/deploy/runtime/defaults/DSLF_Commercial.txt
+++ b/deploy/runtime/defaults/DSLF_Commercial.txt
@@ -1,5 +1,5 @@
 1
-715
+716
 A_aperture
 1
 3
@@ -4205,13 +4205,12 @@ ur_demand_min
 1
 1
 400
-ur_demand_ratchet_percentage
+ur_demand_reactive_power_charge
 1
-2
 1
-12
-5 5 5 5 5 5 5 5 5 5 5 5 
-
+1
+1
+0
 ur_demand_window
 1
 1
@@ -4345,6 +4344,12 @@ ur_has_unused_items
 ur_is_default
 1
 1
+1
+1
+0
+ur_lookback_months
+1
+2
 1
 1
 0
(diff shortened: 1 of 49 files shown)


cpaulgilman:
Should the table be disabled for the "Choose tidal converter from library" option? Otherwise, it's not clear what happens when I import device data from a file (while in library mode): The data in the table updates to the data from the file, but if I reload the page, it reverts to the library selection. The disadvantage of disabling the table is that you can't export data from the library to a file to modify.

And, a separate issue/question:

If I switch from RM1 to RM2, I get a "Size of Power Curve is not equal to Tidal Resource" error. Does the device power curve have to have the same number of bins as the resource table? I thought the model would interpolate if the tables don't match.

mjprilliman:
> Should the table be disabled for the "Choose tidal converter from library" option? Otherwise, it's not clear what happens when I import device data from a file (while in library mode): The data in the table updates to the data from the file, but if I reload the page, it reverts to the library selection. The disadvantage of disabling the table is that you can't export data from the library to a file to modify.
> 
> And, a separate issue/question:
> 
> If I switch from RM1 to RM2, I get a "Size of Power Curve is not equal to Tidal Resource" error. Does the device power curve have to have the same number of bins as the resource table? I thought the model would interpolate if the tables don't match.

On the first question, I wasn't sure if we should disable the ability to modify when in library mode or not. The wave model has the same thing where users can change values in the matrix but reloading the page restores the library value. I'm open to disabling the ability to edit when in library mode but I'll test to see if I can retain the changes on page load first. 

On the second question, at this point no work has been done to interpolate between differently sized arrays for resource and power definitions. We have been leaving it to the user to make sure the lengths of these arrays match. I suppose that could change in the future.

cpaulgilman (commented):
Changing from RM1 to RM2 results in a simulation error. Should the device power curve be modified to have the same number of rows as the default resource data?

image

cpaulgilman (approved):

PR #742, opened 2021-11-09
Sloped Terrain inputs for Single Axis trackers

mjprilliman:
-New terrain tilt angle and terrain azimuth (measured East of North) inputs for subarrays
-Only enabled for 1-axis tracking for this release
-If backtracking enabled, backtracking angle influenced by terrain inputs
-If linear shading is enabled, shadeFraction accounts for terrain input effects on beam self-shading
-Reference: https://www.nrel.gov/docs/fy20osti/76626.pdf


mjprilliman:
[cross_axis_slope_examples.zip](https://github.com/NREL/SAM/files/7508672/cross_axis_slope_examples.zip)

cpaulgilman:
@mjprilliman For which of the following conditions do the terrain slope inputs have an effect on results?

1. Self shading disabled
2. Self shading enabled with linear option
3. Self shading enabled with non-linear option

mjprilliman:
> @mjprilliman For which of the following conditions do the terrain slope inputs have an effect on results?
> 
> 1. Self shading disabled
> 2. Self shading enabled with linear option
> 3. Self shading enabled with non-linear option

1. Only affects results if backtracking is enabled (angle changes to account for slope)
2. Affects results through linear beam portion of self-shading (shadeFraction), can also vary results through backtracking if enabled (eliminates self-shading)
3. Only affects results with backtracking enabled (angle changes to account for slope), which would theoretically eliminate self-shading

cpaulgilman:
This implementation is probably fine for now.

For future improvements:

* Consider enabling slope inputs for combinations that have no effect. For example, disable when self shading is enabled but backtracking is disabled.
* Consider moving to "Shading and Layout" page.

cpaulgilman (approved):

janinefreeman (approved):
I agree with Paul's comments- Matt, can you make issues for those future improvements and assign to patch 1 for now? Thanks!

PR #751, opened 2021-11-11
Add "Copy table >" functionality to P50/P90 

sjanzou:


sjanzou:
SAM project file and P50/P90 weather files for testing  
[SAM_141.zip](https://github.com/NREL/SAM/files/7520353/SAM_141.zip)


with results
image

mjprilliman (approved):
Your example worked well for me. That's a very useful feature!

cpaulgilman:
This works well, but I would recommend changing the label to "Copy table to clipboard" or "Copy to clipboard" and removing the ">" because it suggests that a context menu or other window will open.

sjanzou:
Update button label per comments
image

PR #754, opened 2021-11-11
Batt btm to grid and manual clip

brtietz:
Pairs with SSC PR https://github.com/NREL/ssc/pull/653

Adds UI options to enable disable clip charging and discharging to grid for behind the meter dispatch, and FOM manual dispatch.

- Right now only custom and manual dispatch can discharge to grid from behind the meter. Future PRs may change this.
- Right now only manual and price signals dispatch (BTM) can charge from clipped energy. Future PRs may change this.


deploy/runtime/defaults/PV Battery_Commercial.txt
```diff
@@ -1,5 +1,5 @@
 1
-1058
+1073
```
mjprilliman, line 2: Defaults note: PV Battery Commercial has manual schedule with 4 weekday periods and only 2 periods enabled. Manual dispatch is not the default dispatch mode for PV Battery Commercial

deploy/runtime/ui/Battery Dispatch Automated FOM.txt
```diff
@@ -656,7 +656,7 @@ Battery Dispatch FOM Automated
 TextEntry
 
 0
-2642
+2567
```
mjprilliman, line 659: image
The labels are a bit unclear is charging from clipped power supposed to be allowed for a custom time series?

deploy/runtime/ui/Battery Dispatch Options FOM.txt
```diff
@@ -438,25 +438,24 @@ function show_hide_charge_options()
...
 	{
-		show('batt_dispatch_auto_can_charge',option != 3);
-		show('batt_dispatch_auto_can_gridcharge', option !=3 );
-		show('batt_dispatch_auto_can_clipcharge', option != 3);
-		if (technology() == 'Fuel Cell') 
-		{ show('batt_dispatch_auto_can_fuelcellcharge', option !=3); }
-		else
-		{ show('batt_dispatch_auto_can_fuelcellcharge', false); }
+		enable('batt_dispatch_auto_can_charge',option != 3);
+		enable('batt_dispatch_auto_can_gridcharge', option !=3 );
```
mjprilliman, line 442: image
Fuel Cell Single Owner might need some enable checks

mjprilliman: image
Page load order is not correct for Fuel Cell Single Owner

mjprilliman (commented):
A few questions and UI quirks, but otherwise looks good. I'm trusting ssc to test result changes for different charging and discharging conditions

deploy/runtime/defaults/PV Battery_Commercial.txt
brtietz: Peak shaving as the default dispatch is intentional.
Great point about only having two periods of discharging, I'll get that updated such that it always discharges during the high cost period.

deploy/runtime/ui/Battery Dispatch Automated FOM.txt
brtietz: This is a bug, clipped power should be disabled for custom time series. I'll get this fixed in the next push.

brtietz:
@cpaulgilman Are we intending a battery paired with the fuel cell to be able to discharge to grid in this release? If so, did we want to add those to this PR or a subsequent PR?

I also made some changes to the enable/disable behavior to better reflect the powerflow calculations (which are used by everything, including fuel cell) so this might be worth another look from you.

PR #756, opened 2021-11-11
Update pvsamv1 to apply DC degradation linearly

janinefreeman:
Pairs with NREL/ssc#654 . This updates the user interface, upgrade script, and defaults results. FYI @cpaulgilman this will require a documentation update.


brtietz (changes requested):
Something is wrong with test_script_ow.lk here - why are 146 lines now 46?

janinefreeman:
Oh dear. I'll take a look and fix that when I re-run the script before merging.

brtietz (approved):

PR #757, opened 2021-11-11
Fix bug in billing demand yr 0 peaks

brtietz:
When merging https://github.com/NREL/SAM/pull/728, I missed the fact that setting chk_update_peaks to 1 prevented the callback that set the year 0 peaks from running. Add a check to that variable immediately upon enabling billing demand calcs, so they values are always present.

### DIFF
--- a/deploy/runtime/ui/Utility Rate - Billing Demand.txt
+++ b/deploy/runtime/ui/Utility Rate - Billing Demand.txt
@@ -755,14 +755,15 @@ Electricity Rates - Billing Demand
 MonthlyFactor
 
 0
-2850
+3091
 on_load{'Utility Rate - Billing Demand'} = define() {
 	enable_ratchet_demand_charges();
+	check_update_year_one_peaks(true);
 	firstyear_demand_peaks_message();
 };
 
 on_change{'chk_update_peaks'} = define() {
-	check_update_year_one_peaks();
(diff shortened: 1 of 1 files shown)


brtietz:
Failure matches version of develop used to create this branch. Develop was fixed here: https://github.com/NREL/SAM/runs/4182767103. I can merge in develop and re-run if desired, but refraining for now to save Travis credits.

cpaulgilman (changes requested):
If Use Monthly Peaks is checked, the monthly peaks array should be updated on_load so that if you change the load data the peaks are updated. For example:

1. On Electricity Rates page, check **Use monthly peaks from Year 1...**.
2. On Electric Load page, change **Electric load scaling factor** to 2.
3. Go back to Electricity Rates page and clear Use Monthly Peaks checkbox and click **Edit values* to see values.

brtietz:
@cpaulgilman Good catch! In that case, do we want the yes/no dialogue box to pop up, or should the overwrite happen automatically if chk_update_peaks == 1?

cpaulgilman:
> Good catch! In that case, do we want the yes/no dialogue box to pop up, or should the overwrite happen automatically if chk_update_peaks == 1?

I think it can be automatic.

cpaulgilman (approved):
Looks good now. Thanks!

PR #758, opened 2021-11-11
Add unsupported rate units error message

### LINKED ISSUE #629: Give users notice when the URDB rate download is unsupported in SAM
Right now SAM only supports four types of utility rate units: kWh, kWh/kW, kWh daily, and kWh/kW daily. However, URDB supports additional rate units such as kVA and hp: https://openei.org/services/doc/rest/util_rates/?version=8

If the user attempts to download one of these rates, provide a notice that they may need to do the conversion themselves.

An example rate: https://openei.org/apps/IURDB/rate/view/539fbce2ec4f024c27d8756d#1__Basic_Information

A comment from Steve on a previous PR with more possible examples: https://github.com/NREL/ssc/pull/580#pullrequestreview-671111045

brtietz:
Adds a message about unsupported rate units to the "unused items" dialogue.

Should we be taking a similar approach with demand charge units that are not kW?

Fixes #629

### DIFF
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -3631,7 +3631,7 @@ void fcall_urdb_get(lk::invoke_t &cxt)
 
         // Unsupported units
         if (!rate.EnergyUnits.IsEmpty()) {
-            rate_notes.append(wxString::Format("SAM does not model energy charges with %s units. Energy charges are set with kWh units, which may not represent the actual rate.", rate.EnergyUnits));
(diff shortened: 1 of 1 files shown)


brtietz:
Failure matches version of develop used to create this branch. Develop was fixed here: https://github.com/NREL/SAM/runs/4182767103. I can merge in develop and re-run if desired, but refraining for now to save Travis credits.

cpaulgilman (approved):
Would be good to have this merged into Develop for my work updating to URDB v8.

PR #761, opened 2021-11-12
Fix NSRDB download location name check

### LINKED ISSUE #710: latitude-comma-space-longitude has unspecified behavior
### Describe the bug

SAM's solar resource download accepts latitude-comma-longitude successfully, but adding a space (the default when you copy from Google Maps) converts the lat-lon pair to a different lat-lon or an arbitrary address.

### To Reproduce

1. Open a version of SAM with api-keys (2020.11.29 r2 or a beta)
2. Go to the location and resource page
3. Enter (for example) "39.71, -105.1", select "choose year"
4. See that the lat and lon in the "header data from weather file" do not match what was wntered

### Expected behavior

The lk script responsible for the conversion should strip the spaces for ease of use

### Screenshots

image

cpaulgilman:
Fixes #710 

This fix should allow location name entered as any string, a city/state pair separated by a comma, or a lat/lon pair separated by a comma followed by a space, comma with no space, or space with no comma.


brtietz (approved):
The lk changes had the intended effect, I can now copy-paste a lat/lon pair from google maps and it sends the right thing to NSRDB. However, I'm getting the following error from NSRDB afterwards:

image

Should that be fixed here or in a different PR?

cpaulgilman:
The error message about "required 'email' parameter" should only happen when you test on a non-NREL version of SAM because the email value comes from the registration process. We can double check this when we build a Beta version.

Issue #458, opened 2020-11-24
LK small numbers turned to zeros


dguittet:
In LK callbacks, anything after 6 decimal places is not being included when writing the variable from ssc_exec().

https://github.com/NREL/SAM/pull/454

mjprilliman:
I believe this issue also affects very large numbers as well. We seem to be losing for 9 digits (hundred million) in front of the decimal point. And the printed number is not matching up with what is being used for calculations

dguittet:
Could you please post a sample LK script?

cpaulgilman:
```
x = 1.0834e-19;
outln(x);

x = 1.0834e-7;
outln(x);

x = 1.0834e-6;
outln(x);
```

mjprilliman:
![image](https://user-images.githubusercontent.com/54449384/100682977-98372980-3334-11eb-9115-9b5a316bd786.png)
![image](https://user-images.githubusercontent.com/54449384/100683014-aab16300-3334-11eb-8331-fc712130043b.png)
![image](https://user-images.githubusercontent.com/54449384/100683056-c3ba1400-3334-11eb-86c5-4be06aa12a71.png)






You can mess around with the numbers on that. Comment out the lines on two and unselect the checkmark on 'Override cost structure' to only focus on 1. I can try to come up with better use cases tomorrow. 
[LK_number_testing.zip](https://github.com/NREL/SAM/files/5619583/LK_number_testing.zip)
The number that is printed seems to vary based on what the last two digits are with no clear pattern. Not entirely sure what is going on it may be a different issue. I tried the different number formats on the UI object and it didn't work outside of 
[LK_number_testing.zip](https://github.com/NREL/SAM/files/5619580/LK_number_testing.zip)

sjanzou:
Issue with the lkscript "set" function fixed in SAM_458 branch of SAM - pull request #763 

Issue with small numbers in "outln" lkscript is fixed in SAM_458 branch of lk - pull request https://github.com/NREL/lk/pull/25

Issue with values from numeric control passed through ssc_exec in SAM_458 branch of wex - pull request https://github.com/NREL/wex/pull/125

sjanzou:
This is a time intensive task that will be reolved manually by changing all calculated variables to Real with fixed decimals per discussion at https://github.com/NREL/SAM/pull/763#issuecomment-969394134

PR #763, opened 2021-11-13
Fix first issue with lk "set" function in #458

sjanzou:


cpaulgilman (approved):
Looks good. I tested the small numbers LK example and the ME Wave LK script.

mjprilliman:
image

Did this affect the decimal reporting in calculated variables?

sjanzou:
> image
> 
> Did this affect the decimal reporting in calculated variables?

The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.

cpaulgilman:
> > image
> > Did this affect the decimal reporting in calculated variables?
> 
> The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.

Is a solution to change the definition of the "Generic" format so it displays up to four decimal places?

sjanzou:
> > > image
> > > Did this affect the decimal reporting in calculated variables?
> > 
> > 
> > The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.
> 
> Is a solution to change the definition of the "Generic" format so it displays up to four decimal places?

Possibly, then we still need another format for the inputs for which we want to keep as much precsision as possible (e.g. user entered 6 paramter modules). Doubles have about 16 digits of precision according to most sources.

cpaulgilman:
> > > > image
> > > > Did this affect the decimal reporting in calculated variables?
> > > 
> > > 
> > > The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.
> > 
> > 
> > Is a solution to change the definition of the "Generic" format so it displays up to four decimal places?
> 
> Possibly, then we still need another format for the inputs for which we want to keep as much precsision as possible (e.g. user entered 6 paramter modules). Doubles have about 16 digits of precision according to most sources.

Otherwise, do we write a utility to convert all variables with "Generic" format to "Fixed" with 3 or 4 decimals, and then manually change the ones that need to be Generic for maximum precision?

sjanzou:
> > > > > image
> > > > > Did this affect the decimal reporting in calculated variables?
> > > > 
> > > > 
> > > > The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.
> > > 
> > > 
> > > Is a solution to change the definition of the "Generic" format so it displays up to four decimal places?
> > 
> > 
> > Possibly, then we still need another format for the inputs for which we want to keep as much precsision as possible (e.g. user entered 6 paramter modules). Doubles have about 16 digits of precision according to most sources.
> 
> Otherwise, do we write a utility to convert all variables with "Generic" format to "Fixed" with 3 or 4 decimals, and then manually change the ones that need to be Generic for maximum precision?

Hmmm... maybe we can start a poll? I think some may not like 4 fixed digits for many of the large calculated values... I think the maximum precision is a much smaller subset and thaose can be updated manually during loc down...

cpaulgilman:
> > > > > > image
> > > > > > Did this affect the decimal reporting in calculated variables?
> > > > > 
> > > > > 
> > > > > The wex fix to the wxNumeric control so that more precision can be sent through ssc_exec caused this issue.
> > > > 
> > > > 
> > > > Is a solution to change the definition of the "Generic" format so it displays up to four decimal places?
> > > 
> > > 
> > > Possibly, then we still need another format for the inputs for which we want to keep as much precsision as possible (e.g. user entered 6 paramter modules). Doubles have about 16 digits of precision according to most sources.
> > 
> > 
> > Otherwise, do we write a utility to convert all variables with "Generic" format to "Fixed" with 3 or 4 decimals, and then manually change the ones that need to be Generic for maximum precision?
> 
> Hmmm... maybe we can start a poll? I think some may not like 4 fixed digits for many of the large calculated values... I think the maximum precision is a much smaller subset and thaose can be updated manually during loc down...

True, would 3 be better? Some of the inputs already have a fixed format. If we made this change, I think we could plan to clean up number formats over time.

PR #765, opened 2021-11-14
Stretch multi hour gen outputs to 8760 array for CombineCases

mjprilliman:
-Check for array lengths that are multiple of 8760 (ex. 2920) to stretch out into 8760 gen array of kW values for Combine Cases
-Designed for MEwave, should be flexible for any uniform arrays of > 1 hour
-Should this be extended to include sub-hourly?

### DIFF
--- a/deploy/runtime/defaults/DSLF_Commercial.txt
+++ b/deploy/runtime/defaults/DSLF_Commercial.txt
@@ -1,5 +1,5 @@
 1
-715
+716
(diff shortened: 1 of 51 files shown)


cpaulgilman (approved):
Seems to work for Generic-Battery combining Wind and ME Wave, so I think this is good.

We can consider handling subhourly in the future based on actual use cases as we get feedback from users.

PR #766, opened 2021-11-15
Standalone battery models in SAM

mjprilliman:
-Battery model without power generation component
-Specific cost inputs UI forms, dispatch option rules
-Need to set defaults for standalone battery cases (BTM and FOM)


brtietz:
Great work! Some notes/changes:

In our meeting on Friday, we had discussed setting up a dialogue box if the user attempts to set up a custom weather file. I get this error currently:

> Could not evaluate callback function:batt_select_weather_file_forecast_peak_shaving->on_change
[27]: eval error in statement list
[27]: error inside function call
[28]: eval error in statement list

Two bugs on the summary/metrics page for standalone-commercial: discounted_payback rather than "discounted payback period" and the second box is empty:

image

Also bugs in the metrics for FOM:

image

Do we want to be able to simulate grid outages with a stand-alone battery? The outages would have to be shorter than a pv-battery or generic-battery system, but it still could be a useful feature, especially when comparing to the other two systems. (apologies if we already made this decision and I forgot about it)

I think defaults for this can be done in a subsequent PR (with the rest of the defaults for this release). One note though: for commercial peak shaving, a 100 kW, 400 kWh system gives a positive NPV.

cpaulgilman:
I'm also seeing those missing output variables, including battery replacement cost in the cash flow:

image

Maybe change the short name in startup.lk from "Standalone-Battery" to "Battery" to give a little more space here:

image

cpaulgilman:
Also need message box to prevent choosing DC-connected battery:

image

mjprilliman:
> Also need message box to prevent choosing DC-connected battery:
> 
> image

@cpaulgilman @brtietz The missing outputs variables issue seems to happen when NaN is returned from the compute module.  For example in FOM cases an IRR Do we have a current fix for this or do we just need to work on defaults that return a IRR or payback period that works?

cpaulgilman:
> The missing outputs variables issue seems to happen when NaN is returned from the compute module. For example in FOM cases an IRR Do we have a current fix for this or do we just need to work on defaults that return a IRR or payback period that works?

The behavior in SAM 2020.11.29 r2 is to show the variable label in the Metrics table even when the value is NaN, which is what we want:

image

mjprilliman:
@cpaulgilman @brtietz I will have to dig into that issue in a little bit. The outputs are filling correctly in the Data Tables (with NaN) but not in metrics.lk

mjprilliman:
image

This issue is not exclusive to Standalone Battery. I believe the label is not showing when the metric() function is used in metrics.lk as opposed to metric_table() and metric_row(). These values are used in Host Developer configs which properly shows NaN values: 

image

@cpaulgilman @sjanzou any ideas?

Issue #670, opened 2021-07-28
Variables with commas in label throw off parametric exports to csv


mjprilliman:
### Describe the bug

Parametric exports to csv misalign columns when there is a comma in the variable title

### Expected behavior

Either commas cannot be included in variable labels or checking needs to occur in the export to csv

### Screenshots
![image](https://user-images.githubusercontent.com/54449384/127359623-3b9fb3e4-80e4-4b12-a944-bbf928b357be.png)

![image](https://user-images.githubusercontent.com/54449384/127359722-bf78a571-e526-4717-a7f6-0560750d9485.png)

### Operating System and Version

 - OS: [Windows, macOS, Linux]
 - Version [e.g. SAM 2020.2.29 r3]

cpaulgilman:
@mjprilliman I think commas should be avoided in variable labels. I will clean up these BOS variables for the PV, HCPV, PV Battery and Fuel Cell cost UI forms, but there may be other instances on other forms.

cpaulgilman:
For fuel cell costs, clean up current

![image](https://user-images.githubusercontent.com/27711933/128416861-d567bb33-d4f0-47a3-89bc-c7c47834c2d9.png)

to look like this:

![image](https://user-images.githubusercontent.com/27711933/128416881-3e72b726-38a1-4bb0-8b32-5334d9f2c6f4.png)

sjanzou:
Latest test case 
![image](https://user-images.githubusercontent.com/6498311/141751126-246c66c5-90d8-42c9-9d20-0a9a11f28a96.png)

sjanzou:
With pull request 767, test case results in exported csv:
![image](https://user-images.githubusercontent.com/6498311/141756503-bc81a489-5569-4f6b-972b-b36686abe2c5.png)

Test case attached here:
[SAM_670_test.zip](https://github.com/NREL/SAM/files/7537432/SAM_670_test.zip)

PR #767, opened 2021-11-15
Fix exporting to csv with labels that have commas SAM #670

sjanzou:


mjprilliman (approved):

cpaulgilman (approved):
Enclosing variable labels in quotes is a good solution for Excel and other software that ignores commas inside quotes for displaying comma-separated data.

Issue #685, opened 2021-08-17
Code generator outputs no longer align with metrics table


sjanzou:
![image](https://user-images.githubusercontent.com/6498311/129694138-37a891c6-487e-4e3a-8504-5b6771b9e17a.png)

Discovered while fixing issue #673

mjprilliman:
@sjanzou #558 May fit here as well?

sjanzou:
With pull request #768 the code generator outputs now align
![image](https://user-images.githubusercontent.com/6498311/141780410-9a835121-ed4c-45ae-bb88-1da739d23e66.png)

SAM project file used to test both #558 and #685 
[test_SAM_558.zip](https://github.com/NREL/SAM/files/7538591/test_SAM_558.zip)

Issue #558, opened 2021-03-09
Generate Code does not recognize metric_table function


mjprilliman:
**Describe the bug**
Generating code from a SAM case throws a callback error on Metrics when it encounters the metric_table function used to mark unique tables on the Summary page

**To Reproduce**
1. Open a ME Wave case
2. Click Generate Code and generate a JSON for inputs
3. See error below
4. Resulting json output file does not generate any metrics or metric labels after the metric_table command

**Screenshots**
![image](https://user-images.githubusercontent.com/54449384/110503597-96120300-80b9-11eb-9f73-83ca57bdb52c.png)

sjanzou:
Working with all inputs and outputs listed in Data Tables - note that some in ME configurations are not computed in SDKTool runs since the <NULL> outputs in the lk script are not listed as compute module inputs or outputs and only calculated in equations
![image](https://user-images.githubusercontent.com/6498311/141779398-8c1db1a7-621d-44c0-98bd-770d4ec0660b.png)

sjanzou:
SAM project file used to test #685 and #558
[test_SAM_558.zip](https://github.com/NREL/SAM/files/7538590/test_SAM_558.zip)

PR #768, opened 2021-11-15
Fix code generation with new multiple metric tables #558 and #685

sjanzou:


mjprilliman (approved):

cpaulgilman (approved):
Looks good. This fixes the UI callback error, but LK for SDKtool still results in null values for metric table values calculated in metrics.lk instead of SSC, which I think is a minor issue.

Issue #637, opened 2021-06-11
Stochastic simulations cause crash after canceling


cpaulgilman:
### To replicate:

1. Open stochastic-crash.sam (created in SAM 2020.11.29): [stochastic-crash.zip](https://github.com/NREL/SAM/files/6167655/stochastic-crash.zip)
2. Click **Stochastic**.
2. Click **Run simulations** and click **Cancel** before simulations finish.
3. Click **Run simulations**.

sjanzou:
With SAM pull request  #774 and wex pull request 127 https://github.com/NREL/wex/pull/127

Cancel when preparing, no crash when run simulations again
<img width="1137" alt="image" src="https://user-images.githubusercontent.com/6498311/141954856-151b6170-849e-4557-aae4-6ec5a914e76a.png">

Cancel when threads are running and no crash when run simulations again
<img width="1388" alt="image" src="https://user-images.githubusercontent.com/6498311/141954768-080bbb70-5b9b-4033-8ec3-0175bad14006.png">

PR #774, opened 2021-11-16
Fix stochastic crash when cancelling and rerunning simulations SAM#637

sjanzou:


mjprilliman (approved):
The test file from #637 worked for me when using the SAM and wex branches.

cpaulgilman (approved):

Issue #504, opened 2021-01-21
JSON for defaults and forms


sjanzou:
RapidJSON replacement of Jsoncpp in ssc and wxJSON in wex.
Save defaults as JSON with calculated values as -777 and indicators as -888

See save_as_JSON branches of lk, wex, ssc and SAM for current progress

sjanzou:
Defaults are saved as JSON. Indicators do not work with equations.

Next step here is forms as JSON

Issue #486, opened 2020-12-18
Use RapidJSON through SAM code


dguittet:
Currently there are at least two different JSON readers in SAM: JsonCpp and WxWidgets'. RapidJSON is much faster: https://github.com/miloyip/nativejson-benchmark

SAM's code base should be modified to use just RapidJSON.

brtietz:
@sjanzou This can be closed, right?

PR #782, opened 2021-11-18
Replace JSONcpp with RapidJSON

sjanzou:
run export_config, defaults, ssc tests and landbosse


sjanzou:
Partially addresses issue  #504 and #486

sjanzou:
goes with ssc pull request 664 https://github.com/NREL/ssc/pull/664

cpaulgilman (approved):
I tested some of the SAM UI functions that use JSON, including URDB downloads, NSRDB, Cambium, marine wave resource, and all are working. These are still using wxWidgets JSON tools. Perhaps we should not change these until a later update.

sjanzou:
> I tested some of the SAM UI functions that use JSON, including URDB downloads, NSRDB, Cambium, marine wave resource, and all are working. These are still using wxWidgets JSON tools. Perhaps we should not change these until a later update.

@cpaulgilman, that is correct... this pull request only replaces JSONcpp with RapidJSON and the SAM dependencies for this are in the landbosse and pythonhandler code used. We are now down to wxJSON and RapidJSON with this pull request and will eliminate wxJSON in another pull request.

PR #804, opened 2021-11-24
Added GE inverter derate curve for voltage 1300

### LINKED ISSUE #803: Update default inverter temperature derate curves for large inverters
The current default inverter temperature derate curves cover the range 390 to 800 Vdc. If possible range should cover small residential to large utility-scale inverters, so perhaps add a curve for a higher voltage than 800.

See here for links to inverter datasheets with temperature derate information: https://github.com/NREL/ssc/issues/442

See here for a discussion of the issue: https://sam.nrel.gov/forum/forum-general/3521

mjprilliman:
## Description

Added GE inverter derate data for inverter with higher operating voltage (Vdc). Added this curve to existing inverter derate curves. 

Source: https://www.ge.com/renewableenergy/sites/default/files/related_documents/lv5plus-solar-inverter-datasheet.pdf

Fixes #803 
See #https://github.com/NREL/ssc/issues/442

## Type of change

## Checklist:

### DIFF
--- a/deploy/runtime/ui/Financial Solution Mode Single Owner.txt
+++ b/deploy/runtime/ui/Financial Solution Mode Single Owner.txt
@@ -523,7 +523,7 @@ MultilineText
 1
 
 
-2169
+2245
 // This form is a copy of Financial Solution Flip Leaseback
 // so ppa_soln_mode radio button can have different labels
 // to indicate IRR target is for the tax investor for flip 
@@ -532,8 +532,9 @@ MultilineText
 on_load{'Financial Solution Mode Single Owner'} = define()
 {
 	soln_mode_ctrl_refresh();
-	if (technology() == "PV Battery" || technology() == "Generic Battery" 
-	|| technology() == "Standalone Battery" || technology() == "Fuel Cell") {
+	if ((technology() == "PV Battery" || technology() == "Generic Battery" 
+	|| technology() == "Standalone Battery" || technology() == "Fuel Cell") &&
+	financing() != "Host Developer") {
 		check_auto_ppa();
 	}
 	display_message();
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
Looks good to me.

One question: Should versions.lk overwrite the old temperature derate table with the new one? I think it should only do that if the old table has default values, i.e., was not modified by user -- not sure if it's worth the effort to code that comparison.

dguittet:
I think this combination of two different inverter derate curve specs (1 high voltage line from 1 high voltage inverter and the rest from a mid-voltage inverter) looks kind of odd. My intended but missed goal for the original example was to show how to input multiple lines for different voltages, not really to provide a single "default" set that would apply to most situations since I don't know what that would even be. I'd propose something like the following for the default since it shows the capability, kind of extrapolates the behavior of the high-voltage inverter, and very often has little impact so the users won't get anything they weren't anticipating.

image

dguittet:
Well... I just made up the 390 V curve based on the 1600 V one. :stuck_out_tongue_winking_eye:

cpaulgilman:
> > Well... I just made up the 390 V curve based on the 1600 V one. 😜
> 
> Aah I understand now I should've read your first comment more closely. I'm not sure what the best path forward would be. @cpaulgilman should we consider having different defaults for the different default system sizes?

I don't think there is a perfect quick solution for a last minute fix before the release. The user-reported issue on the forum was caused when the user chose an inverter with VDCmax greater than the 800 VDC maximum value in the temperature derate table -- that would not be solved by having different default values. 

One solution would be to display a message when that happens like "Maximum inverter DC voltage is greater than the highest voltage in the temperature derate table. See the inverter manufacturer datasheet for information about temperature derating and change the derate table accordingly."

Some thoughts:

1. Ideally, the inverter library would include inverter temperature derate data. 
2. We could write a callback to set the default temperature derate curve when the library choice changes (or when the maximum DC voltage changes so it works with the datasheet option), perhaps for three or four ranges of inverter maximum rated DC voltage. 
3. Darice's made up curves probably work well, but are hard to justify. The current default curves, which I think are from https://files.sma.de/downloads/Temp-Derating-TI-en-15.pdf for smaller inverters has a smaller slope at high temps than the larger inverters, and the made up curves do not reflect that.

dguittet:
I'm not sure how commonly reported this inverter derate data is. If we were to set up something as a default for all inverters, I would go the route of one of minimal complexity and with minimal effect on the simulation. Since it's there as an illustration not as a meaningful input. Just my 2 cents though.

mjprilliman:
Changing the default to one inverter curve to cover wide voltage range with details on how to enter multiple curves available in help

Reference: https://www.ge.com/renewableenergy/sites/default/files/related_documents/lv5plus-solar-inverter-datasheet.pdf

PR #817, opened 2021-11-30
Update units on grid outage page to make it really obvious 0 is grid on

brtietz:
We had a review meeting for the grid outage features this morning, and the main feedback was that people have a strong association with 0 as "off" so we need to make our units convention for the grid outage (1 is outage, grid is off) more obvious in the UI.

I've tried to do that here by adjusting the units in the outage input. I'm open to additional wordsmithing and formatting, but think we need something in addition to 0/1 to make this more obvious.

### DIFF
--- a/deploy/runtime/ui/Grid Outage.txt
+++ b/deploy/runtime/ui/Grid Outage.txt
@@ -30,12 +30,12 @@ Mode
 1
 Label
 5
-3
-0/1
+20
+0=no outage/1=outage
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
Looks good. I committed a couple of changes to revise the variable units label and text in Data Array widget, and also fixed some missing UI variable group labels for variables in BTM PV Battery configs.

PR #822, opened 2021-12-01
Set default Heliostat Availability Constant loss to 1% for all power …

### LINKED ISSUE #50: Power tower heliostat availability not used in simulations

cpaulgilman:
Fix #50


tyneises (approved):
I updated the test results for the new default value. Looks good.

PR #834, opened 2021-12-01
Revised checks for IRR target mode / automated dispatch to use en_batt

mjprilliman:
-Checking the en_batt is assigned and that battery is enabled (en_batt == 1)
-Still checking that Host Developer financial model is not being used as there is not automated dispatch for this model

## Type of change


brtietz (approved):
Looks great, thanks!

PR #843, opened 2021-12-07
Fix broken link html in changelog entry for 2021.12.02

kandersolar:
Congrats on the new SAM release!  There's a small html issue in the changelog causing a link to span several lines of text.  See the last bullet under "Detailed Photovoltaic" through to the first bullet of "Battery Storage":

image


brtietz (approved):
Good catch, thank you!

PR #844, opened 2021-12-07
update sales tax percent to be the same for all techs

brtietz:
# Pull Request Template

## Description

Set all of the sales tax defaults to 5%, update test script accordingly


cpaulgilman (approved):
Defaults for the "Retired,none" configuration are included here. No harm done, but I wonder if that configuration needs a defaults file with variables in it?

PR #845, opened 2021-12-08
GitHub actions for Linux on sam

brtietz:
This pull request mimics the Linux functionality on travis, except that it runs Ubuntu 18.04, and (once merged) will run for every branch.

Future PRs will store the test results as artifacts, and expand to cross platform runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,21 +73,21 @@ jobs:
       with:
         path: lk
         repository: NREL/lk
-        ref: ${LK_BRANCH}
+        ref: ${{ env.LK_BRANCH }}
    
     - name: Checkout wex   
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


sjanzou (approved):
Everything looks good; however, as you mention Ubuntu 18.04 with gcc 7.5.0 and libc 2.27.
We are constrained to CentOS7, gcc 4.8.5 and libc 2.17:
image

brtietz:
We discussed this a bit at the SAM meeting last week. Here are the options for GitHub hosted runners: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners I view this PR as a partial replacement for Travis-CI.

The options for self hosted runners are much broader: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners but we would have to provide a machine.

PR #855, opened 2021-12-10
Batt dispatch choice variables matching default page loads

### LINKED ISSUE #846: PV Battery / Merchant Plant default state of Battery Dispatch page incorrect
To replicate, create a new PV-Battery / Merchant case and go to the Battery Dispatch page. Need to check all FOM battery configurations:

image

Correct state is:

image

Workaround is to change the dispatch option to a different one and then back to Manual Dispatch.

mjprilliman:
## Description

Updated all detailed battery dispatch pages to have the same UI choice, batt_dispatch_choice value sent to ssc, and batt_dispatch_excl variable for correct page load based on default choices. 

Fixes #846 

## Type of change

### DIFF
--- a/deploy/runtime/defaults/Fuel Cell_Single Owner.txt
+++ b/deploy/runtime/defaults/Fuel Cell_Single Owner.txt
@@ -4739,7 +4739,7 @@ ui_batt_life_model
 1
 1
 1
-0
+1
 ui_electricity_rate_option
 1
 1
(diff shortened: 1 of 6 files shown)


cpaulgilman (changes requested):
@mjprilliman This looks good and fixes #855, but I found the default value of `ui_batt_life_model` is incorrect for several configurations. I think the default value should be `ui_batt_life_model=1` for all configurations because the default battery type is Li-ion NMC/Graphite.

I tested the following configurations, but the defaults manager shows others, such as Fuel Cell / Single Owner have the wrong default value for `ui_batt_life_model`

- PV Battery / Merchant
- Generic Battery / Leveraged Partnership Flip
- Generic Battery / All Equity Partnership Flip
- Generic Battery / Sale Leaseback

cpaulgilman (approved):
Looks good. Thanks for making those changes with this PR.

PR #863, opened 2021-12-17
Default diff notebook

brtietz:
Adds ipython notebook to generate diffs between the files in api/api_autogen/library/defaults.

### DIFF
--- a/test/run_defaults_diff.py
+++ b/test/run_defaults_diff.py
@@ -1,51 +1,73 @@
-import json
-import os
-import tempfile
+from PySAM.PySSC import PySSC
 import requests
-import tarfile
-import shutil
-from collections.abc import Iterable
 from collections import OrderedDict
+from collections.abc import Iterable
+import tempfile
+import os
+import sys
+import json
+import tarfile
+from ctypes import *
 
-#################################################################################################################
-#
-# This file compares the SSC Compute Modules input variables and the SAM defaults for these input variables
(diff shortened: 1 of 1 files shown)


dguittet:
I modified the output to be a JSON containing 4 nested dictionaries: 
   1. 'Cmods with new defaults files'
       Cmods listed here have new default configuration files in the new Release. This may occur with a new SAM configuration, or if an existing SAM configuration added or switched a cmod, e.g. PVWatts using v8 instead of v7

   2. 'Cmods with removed defaults files'
        Cmods listed here have deleted default configuration files in the new Release. This may occur with a deleted SAM configuration.
   3. 'Cmods with modified variables'
          There are three categories here: 'Added variables', 'Removed variables' and 'Type changed'.
   4. 'Configs with modified defaults'
          This is an option that records the default values that have changed. For variables whose lengths are greater than or equal to 8760, the value changes are only reported for the first index to save space.

Note: Since this file uses the Default files to search for variable changes, and Default files do not exist for cmods not in a SAM configuration (e.g. modules providing helper functions or older versions, listed here: https://nrel-pysam.readthedocs.io/en/master/Models.html#other-modules-names-and-descriptions), for these cmods, variable changes will not be tracked by this current method.

dguittet:
Let's evaluate adding this to the Github Actions for this PR. If we decide not to add to GHA, then feel free to merge this after you've reviewed the changes I made.

PR #870, opened 2022-01-05
Patch to develop for travis removal

brtietz:
Patch to develop


brtietz:
Need to do this locally due to conflicts

PR #892, opened 2022-01-20
Make JPD file from downloaded time series wave resource data

mjprilliman:
## Description

Incorporating a requested for the wave model. Time series wave resource data is converted to a joint probability distribution (frequency of resource) file on download from the Wave API or on button press in the UI. 

## Type of change

- [ ] This change requires a documentation update (maybe)


cpaulgilman (changes requested):
1. When I select an item in the library and click **Make JPD File**, nothing seems to happen. I expected it to convert the data and prompt me to save a JPD file, or tell me where it saved the JPD file.

2. When I download a time series wave file (for 46.2,-124.2 in this case), SAM seems to communicate with the server, but I get this message and no file is added to my SAM Downloaded Weather Files folder:

image

3. If I click **Cancel** in the download window, I get this UI callback error:

```
Could not evaluate callback function:btn_wave_download->on_change
[314]: reference to unassigned variable:fn
[314]:failed to evaluate function call argument number 0, value = fn
[289]: eval error in statement list
[289]: error inside function call
[146]: eval error in statement list
```

4. The "SAM Downloaded Device Files" is the default folder for Wave Data Files. I think it should be SAM Downloaded Weather files, unless we want to create a new folder wave resource files.

image

5. When I click **Cancel** from the Folder Settings window, I get this message:

image

Let me know if you'd like my help on any of these.

mjprilliman:
> 1. When I select an item in the library and click **Make JPD File**, nothing seems to happen. I expected it to convert the data and prompt me to save a JPD file, or tell me where it saved the JPD file.
> 2. When I download a time series wave file (for 46.2,-124.2 in this case), SAM seems to communicate with the server, but I get this message and no file is added to my SAM Downloaded Weather Files folder:
> 
> image
> 
> 3. If I click **Cancel** in the download window, I get this UI callback error:
> 
> ```
> Could not evaluate callback function:btn_wave_download->on_change
> [314]: reference to unassigned variable:fn
> [314]:failed to evaluate function call argument number 0, value = fn
> [289]: eval error in statement list
> [289]: error inside function call
> [146]: eval error in statement list
> ```
> 
> 4. The "SAM Downloaded Device Files" is the default folder for Wave Data Files. I think it should be SAM Downloaded Weather files, unless we want to create a new folder wave resource files.
> 
> image
> 
> 5. When I click **Cancel** from the Folder Settings window, I get this message:
> 
> image
> 
> Let me know if you'd like my help on any of these.

Thanks for the feedback Paul. Notes below: 

1. I added a msgbox describing where the file was saved. Right now I have the process automated to save the jpd file in the same folder as the time series file. Do you think it should be a file save prompt instead?
2. I think this been fixed in the latest commits. Let me know if you still have issues as I am not able to replicate them. 
3. Same as 2. 
4. I am unsure how these defaults are generated. It seems to have something to do with the 'wave_data_paths' but I do not see it hard-coded anywhere. 
5. What should the behavior be. The solar Folder Settings buttons have the same behavior.

cpaulgilman (approved):

mjprilliman:
Added bug fixes, label updates to this PR:
-Change sidebar label titles to match marine energy terminology
-Fixed label disappearing on ME Array page
-Changed blue bar wave resource switch to radio choice exclusive header page (label could use review)

cpaulgilman (approved):
Looks good. I revised the text. I will update Help before we release Patch 1.

PR #900, opened 2022-01-25
ETES: remove tax incentive and switch financing debt mode

tyneises:


tyneises:
This pull request updates ETES defaults for Patch 1 as follows:
- set federal ITC to zero from previous default of 26%, which was taken from solar default cases
- switch to DSCR method for project debt and set max debt fraction to 60%
- assign a capacity payment of $150,000/MW and credit 100% of nameplate
- Set PPA price to $0.075/kWh to get a default solved IRR of around 7%

PR #903, opened 2022-01-28
Update NSRDB Download Script

janinefreeman:
This PR updates the Sample Script to help users download NSRDB files (NOT the UI callback, but the separate LK script). It adds further error checking and enables the script to name files using a "name" column in the CSV.


cpaulgilman (approved):
Looks good! I added a CSV file to use for testing.

Issue #862, opened 2021-12-17
Default PPA price for CSP models too high


cpaulgilman:
In the process of switching from default Specify IRR target mode to Specify PPA price mode for SAM 2021.12.02, we forgot to change the default PPA price for CSP models. The default PPA price should be approximately equivalent to the PPA price that would result from the defaut IRR target.

@tyneises will check with the CSP team to see what an appropriate PPA price would be. Note LBL's "Utility-scale Research" page has a spreadsheet with trough and tower PPA prices over 2008 - 2021 on the "PPA Price for CSP Plants" worksheet, see the link to Utility-Scale Solar 2021 Edition Data File (XLSM) at the bottom of the page: https://emp.lbl.gov/utility-scale-solar

![image](https://user-images.githubusercontent.com/27711933/146579056-f27ea0be-7397-4724-a700-77361fc19a6e.png)

tyneises:
closed in #904

PR #904, opened 2022-01-28
Update CSP financial defaults

tyneises:
This commit updates CSP financial defaults for the PPA financial models. The 2021.12.02 release switched the financial solution mode from specifying an IRR target and calculating the PPA price to specifying the PPA price and calculating the IRR target. We did not review the default PPA input when we made this switch, and the release default value of 13 cents/kWh results in unrealistically large IRRs. For each CSP technology, we reset the first year PPA price to a value that results in an IRR close to 7%. We also set the default Buy Rate on the Electricity Purchases page to the first year PPA price, although this input is not used in the default case. Similarly, we set the default IRR target to 7%, although this input is not used in the default case.

The release also allowed the default DSCR debt method to calculate unlimited project debt, which caused aggressive/optimistic cases to have very large IRRs as the debt fraction approached 100% and an undefined IRR as the debt fraction grew to >= 100%. For all CSP cases, we set the default maximum debt fraction in the DSCR method to 60%.

Finally, for the "Partnership Flip with Debt" and "Partnership Flip without Debt" PPA financial models, we changed the IRR target year in the Physical Trough, Molten Salt Power Tower, and Generic CSP model to 9 years to be consistent with these inputs in other technology models.


tyneises:
FYI @cpaulgilman

tyneises:
closes #862

PR #905, opened 2022-01-28
updated capacity factor labels to include DC

janinefreeman:
Updates labels for PV capacity factors in the metrics tables to include the word "DC" to distinguish them from AC capacity factors.


cpaulgilman (changes requested):
Looks good. Please change the metrics table label for HCPV from "capacity factor" to "DC capacity factor."

cpaulgilman (approved):

PR #910, opened 2022-01-31
Geothermal Drilling cost curve updates

mjprilliman:
## Description

-Update drilling and injection cost curves to be based off scenario (4 options), well diameter (2 options), and well type (2 options)
-Add plot showing where along the cost curve scenarios the selection lies
-interface redesign

### DIFF
--- a/deploy/runtime/defaults/Geothermal Power_Single Owner.txt
+++ b/deploy/runtime/defaults/Geothermal Power_Single Owner.txt
@@ -1,5 +1,5 @@
 1
-611
+610
 CT
 1
 1
@@ -2866,12 +2866,6 @@ ppa_soln_mode
 1
 1
 1
-ppi_base_year
-1
-1
-1
-1
-0
 prop_tax_assessed_decline
 1
 1
--- a/deploy/runtime/ui/Geothermal Costs.txt
+++ b/deploy/runtime/ui/Geothermal Costs.txt
@@ -1,7 +1,7 @@
 Geothermal Costs
-824
-1863
-134
+817
+1814
+132
 GroupBox
 
 1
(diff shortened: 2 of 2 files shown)


mjprilliman:
@cpaulgilman I could use some interface design notes on this

cpaulgilman:
@mjprilliman This is pretty cool. Nice work! I can make  couple of small tweaks to make the layout clearer if you don't mind. 

A couple of questions:

1. Do you have a preference for the "Update Plot" button vs automatically updating it based on on_change events for the cost curve, well type, and well diameter options? It seems like it would be best for the state of the plots to always match the state of the options and calculated dollar amounts. But maybe I'm missing a reason to have the user update the plot manually?

2. This will need documentation. Do you have references with definitions of the baseline, intermediate, and ideal cost curves?

mjprilliman:
> @mjprilliman This is pretty cool. Nice work! I can make couple of small tweaks to make the layout clearer if you don't mind.
> 
> A couple of questions:
> 
> 1. Do you have a preference for the "Update Plot" button vs automatically updating it based on on_change events for the cost curve, well type, and well diameter options? It seems like it would be best for the state of the plots to always match the state of the options and calculated dollar amounts. But maybe I'm missing a reason to have the user update the plot manually?
> 2. This will need documentation. Do you have references with definitions of the baseline, intermediate, and ideal cost curves?

1. I think I like updating for any changes too and I don't see a reason not to so I will update that. 
2. I will get a reference from the people that worked on the curves. 
Also I seemed to have caused a configuration to crash so I will investigate that further. I believe it is coming from Geothermal LCOE or Geothermal None

mjprilliman:
Moved to https://github.com/NREL/SAM/pull/912/files

Issue #694, opened 2021-08-26
Send-to-Excel with Equations fails for Single Owner with time series PPA price multipliers


cpaulgilman:
### Single Owner with time series PPA price multipliers

See this forum post for a description and example file: https://sam.nrel.gov/forum/forum-general/3569

Also, https://sam.nrel.gov/forum/forum-general/3718

Verify that solution works for all financial models that use PPA price multipliers.

Additional send-to-Excel with Equations issues to fix:

* Fix sign of Total IBI and Total CBI under investing activities for all spreadsheets.

* Fix LCOS rows in cash flow.

* Land lease operating expense for FOM financial models

* Fix annual energy to grid/from grid/net

* Test for all financial models

* Fix electricity purchases per https://github.com/NREL/SAM/pull/785

PR #919, opened 2022-02-07
Fix cf energy sales

brtietz:
Cashflow.lk changes to go with https://github.com/NREL/ssc/pull/750

Additional changes may be needed for "send to excel with equations" please advise

### DIFF
--- a/deploy/runtime/cashflow.lk
+++ b/deploy/runtime/cashflow.lk
@@ -174,7 +174,7 @@ function cf_revenue()
 	elseif ( fin == 'Single Owner' ) // Community solar does not earn additional project revenue
 	{
 		cfline( 'cf_ppa_price' , -2 );
-		cfline( 'cf_energy_sales_value', 0);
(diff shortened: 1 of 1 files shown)


cpaulgilman:
@brtietz This should be branched off of Patch instead of Develop. Similar to SSC, my om_cost_inputs branch has other fixes related to cash flow, so I would like to merge fix_cf_energy_sales into om_cost_inputs and then into Patch if that's OK by you.

brtietz:
@cpaulgilman looks like it was just a target branch issue in the pull request - I've fixed that (I wonder if we should adjust that default during patch season). Yes, feel free to do that merging route.

cpaulgilman:
Send-to-Excel with equations will be handled when I address https://github.com/NREL/SAM/issues/694

brtietz:
Included in https://github.com/NREL/SAM/pull/927 - closing this PR

PR #920, opened 2022-02-07
Sizing of battery temperature array time series option

### LINKED ISSUE #918: Standalone battery environment temperature options causes simulation error
### Describe the bug

See https://sam.nrel.gov/forum/forum-general/3759 for a description, example file, and workaround solution.

### To Reproduce

1. Create a default Standalone Battery / Commercial case.

2. Go to the Electric Load page, click Edit array, Change time step and type 15. 

3. Run a simulation and note the simulation message.

4. On the Battery Cell and System page, under Battery Thermal, change the environment temp option to Enter single fixed temperature.

5. Run a simulation and note simulation message.

### Expected behavior

A. The default environment temp option for all standalone battery configurations should probably be single fixed temperature.

B. SSC should only use the time series environment temp data when the environment temp option is set to enter time series data, and only generate that error based on the option. The UI always creates an array of battery environment temperatures. For ‘Enter single fixed temperature’, the array is sized then each index is written over with the fixed value. There is a bug where the length of that array is hard coded to 8760 for Standalone battery. Should be fixed to dynamically allocate based on either load or curtailment. 

C. The error message should say "Environment temperature input length must equal number of weather file and/or electric load data records."

SAM 2021.12.02

mjprilliman:
## Description

-Size battery temperature array size based on curtailment array size for Standalone battery FOM, load array size for Standalone battery BTM
-Fix defaults to used fixed battery temperature for all battery configs
-Closes #918 

## Type of change


cpaulgilman (approved):
Looks good!

PR #934, opened 2022-02-12
Fixes SAM issue 828 

sjanzou:


brtietz:
Merge conflicts solved in https://github.com/NREL/SAM/pull/937 Closing- but I won't be offended if this is reopened because I missed something.

Issue #828, opened 2021-12-01
Wind Wizard fails in release candidate


brtietz:
### Describe the bug

Running the wind wizard in the release candidate generates an LK error:

![image](https://user-images.githubusercontent.com/5530592/144294047-85e3e041-0975-4054-87ac-6dc10d1ce97e.png)

### To Reproduce

1. Click through all of the default options in the wind wizard
2. Run the simulation
3. Don't run parametrics
4. Don't output a report (error still occurs if you do)
5. See the above error

### Expected behavior

Script should exit without errors

### Operating System and Version

 - OS: Windows
 - Version 2021.12.02 RC

PR #935, opened 2022-02-12
Manually create pull request to fix SAM #828 - git pull request of patch into develop fails...

sjanzou:


sjanzou:
patch into develop fail...

PR #940, opened 2022-02-17
NMC/Gr bug fix & Switch to cycle_counts vector

dguittet:
Update to PySAM files & SAM test results due to https://github.com/NREL/ssc/pull/763


brtietz:
Looks like this branch ran against ssc patch instead of batt_nmc_life: 
image

I'll have to dig into that tomorrow. The test results changes look reasonable, it might be best practice to merge https://github.com/NREL/ssc/pull/763 into ssc patch first and re-run the tests here.

brtietz (approved):
Changes look good - thanks for re-running the export!

PR #944, opened 2022-02-18
improve the github targeting so we get the correct ssc branch

brtietz:
Previously, pull request builds would use ssc from patch in order to run CI. This sometimes resulted in test failures.

Use a conditional statement to sometimes use GITHUB_HEAD_REF so we use the correct SSC branch, regardless of build type. More details: https://docs.github.com/en/actions/learn-github-actions/environment-variables Helpful SO answer for how the bash if works: https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

CI here should pick up on the improved_ci_target SSC branch for both builds. Failure is expected due to failures on patch, will pull in patch as needed. No rush, this doesn't affect the code in the patch.


brtietz:
Pull request builds now check out the correct branch.

image

dguittet (approved):
Sorry for the delay on this! I lost it in my emails.

PR #956, opened 2022-02-24
2021.r1.ssc268 from patch into develop (improved CI)

brtietz:
Merge patch into develop with correct CI base branch.

We only have to do this additional branch and commit when someone branches from develop instead of patch for their patch work, so we need to keep a closer eye on that during patch PRs.


sjanzou (commented):
Thanks for catching the CI issue!

PR #973, opened 2022-03-07
Add subcooled outlet option to DSG LF IPH model

tyneises:
image


cpaulgilman:
> The new radio button on the System Design page lets users decide between solar field that deliver boiling 2-phase steam (default) or subcooled pressurized water where the outlet remains a single-phase liquid. The outlet temperature of the subcooled case is set relative to the calculated saturation temperature (which in turn is a function of the ‘Heat sink in let pressure’ input) using the new ‘Subcooled dT’ input.

PR #983, opened 2022-03-14
Puerto Rico Mini Grid Wizard 

sjanzou:


sjanzou:
Wizard now scales sizing based on number of participants
image

deploy/runtime/quickstart/PuertoRicoMiniGrid.lk
```diff
@@ -0,0 +1,331 @@
+// Puerto Rico Mini Grid macro using PuertoRicoMiniGrid.sam
```
brtietz, line 1: Nate had requested we change the name of this (including the file name, I think) since it could be used for any mini grid, not just mini grids in Puerto Rico

brtietz (commented):
Scaling factor and outputs look great! My only comment is that I think we ultimately want this to be less location specific prior to releasing it. Happy to discuss further as needed.

cpaulgilman (commented):
Load scaling and graph output looks great. Some comments:

1. LK script file names for wizards should be consistent with other files in quickstart folder so that all files either use CamelCase, title case or sentence case. In general we use sentence case in SAM, but title case could be appropriate here. Also note getting started guide command uses sentence case:

image

2. Text in message boxes could use some editing, but deferring to @n8blair who I think was planning to do some editing.

3. If/when we add weather file download capability, I would recommend storing downloaded weather files in `wfdownloaddir()` instead of `runtimedir() + 'quickstart/puerto_rico_weather'` to improve portability across different SAM versions. This would also make the weather files available for other simulations.

cpaulgilman (approved):
My comments addressed in 7c325e1d082bce87f3af8aa3668e109705b17a19.

brtietz (approved):
Looks good, thanks!

PR #999, opened 2022-04-01
Fix REopt API call functions in UI

### LINKED ISSUE #974: REopt "Get size and dispatch" does not set battery dispatch option correctly
### Describe the bug

Running REopt from the Battery Cell and System page results in values of time series input battery power targets variable being set correctly, but the optimization option on the Battery Dispatch page does not get set to **Input battery power targets**.

### To Reproduce

1. Create a default PV Battery / Commercial case.
2. On the Battery Cell and System page, click **Get size and dispatch**.
3. When optimization finishes, click Yes to change inputs, and go to Battery Dispach page. Note dispatch option is set to the default peak shaving option instead of input battery targets.

### Expected behavior

Callback script should set the dispatch option to the correct value, and provide better information in the message box about what dispatch inputs are changed by the optimization.

Also note incorrect "ReOpt" in some UI elements. Should be "REopt" per https://developer.nrel.gov/docs/energy-optimization/reopt/

Need to check behavior for different BTM configurations with PV Battery and PVWatts Battery.

cpaulgilman:
## Description

Update UI battery sizing variables to support kWh or h option for battery duration implemented in #586.

Fix logic in UI callback so REopt API call results are correctly applied based on user responses to prompts.

Fix "REopt" capitalization in UI text and REopt widgets.

Test for PV Battery, Fuel Cell, and PVWatts Battery configurations.

Goes with SAM-private PR.

Fixes #974

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings

### DIFF
--- a/deploy/runtime/ui/Battery Model Simple with REopt.txt
+++ b/deploy/runtime/ui/Battery Model Simple with REopt.txt
@@ -1,6 +1,6 @@
 Battery Model Simple with REopt
-588
-483
+885
+333
 13
 GroupBox
 
@@ -18,10 +18,10 @@ Y
 0
 Width
 3
-572
+872
 Height
 3
-471
+330
 Tool Tip
 5
 0
@@ -42,7 +42,7 @@ Name
 batt_custom_dispatch
 X
 3
-246
+204
 Y
 3
 138
@@ -79,13 +79,13 @@ Name
 batt_dispatch_label
 X
 3
-42
+39
 Y
 3
-111
+108
 Width
 3
-510
+831
 Height
 3
 24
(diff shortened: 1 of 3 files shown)


sjanzou (approved):
Everything seems to work fine, one question about the PV-Battery/Residential defaults using ReOpt - sizing comes back with 0kW battery size. Should we changed the defaults?
image

cpaulgilman:
@sjanzou I'll check with the team and see if we can come up with a default PV Battery / Residential case that is more cost-effective than PV-only. Note that the NPV for the default case is higher with no storage than with storage because of the cost of the battery, even with the exaggerated fictional rate structure I came up with to get a positive NPV.

PR #1012, opened 2022-04-26
run export config with latest patch data

brtietz:
Run export config and api autogen with the latest changes on patch.


api/api_autogen/library/defaults/Battery_FuelCellCommercial.json
```diff
@@ -169,6 +169,8 @@
 		},
 		"PriceSignal": {
 		},
+		"Revenue": {
```
mjprilliman, line 172: Why is this added to every file?

mjprilliman (approved):
Just one question for my understanding

api/api_autogen/library/defaults/Battery_FuelCellCommercial.json
brtietz: That comes from https://github.com/NREL/ssc/pull/787/files - which is an ssc vartable input group for revenue forecasting.

PR #1014, opened 2022-04-26
Fix Physical Trough bug when number of SCAs in loop is changed

tyneises:
In the Physical Trough model on the Solar Field page, changing the number of SCAs in the Single Loop Configuration results in a bug in the 2022.12.02 release.

To reproduce
- Open default Physical Trough
- Change the number of SCAs in the Single Loop Configuration from 8 to 4
- Run the simulation - SAM should hard crash

image


tyneises:
FYI @cpaulgilman @Matthew-Boyd

PR #1015, opened 2022-04-26
add ptes to pysam

tyneises:


brtietz (approved):
This ran successfully on my machine. I committed the changes from running api autogen and export config

PR #1018, opened 2022-04-26
Improve handling of NSRDB API call errors

### LINKED ISSUE #1017: NSRDB solar resource download returns script error when HTML error encountered
### Describe the bug

When the NSRDB API request returns HTML instead of JSON, UI generates an error:

image

This is difficult to replicate because it requires that the NSRDB API be down.

cpaulgilman:
Some types of API errors return JSON with an error message, and others return HTML with an error code. This fix better handles situation where HTML is returned.

One test is to try to download a file for a location that is not in the NSRDB, for example, this lat/lon pair: 65,-122.7.

Fixes #1017

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have checked my code and corrected any misspellings

### DIFF
--- a/deploy/runtime/ui/Solar Resource Data.txt
+++ b/deploy/runtime/ui/Solar Resource Data.txt
@@ -2624,7 +2624,7 @@ equations{'ui_step_minutes'} = define() {
 	return(${step}/60);
 };
 
-32799
+33231
 on_load{'Solar Resource Data'} = define() {
 	//msgbox('on load');
 	//hide legacy data for 2020.2.29 Patch 3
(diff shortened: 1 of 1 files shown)


sjanzou (approved):
Looks and works great!

cpaulgilman:
> Does anything in nsrdb.cpp need to be adjusted to match for the advanced download? There was a recent error in the psm3-tmy specification that this sort of debugging would have helped with. See #1022 and #1023 for details.

That's a good point. I was able to write the code  in the UI callback script while the NSRDB was down so I could test to make sure it worked properly. I'm hesitant to replicate that code in nsrdb.cpp now that I can't test it with the NSRDB back up and running. Depending on the state of the NSRDB it either returns a CSV weather file, a JSON file with error messages, or an HTML file with error messages.

PR #1023, opened 2022-04-28
update attributes for psm3-tmy to match the lk script

### LINKED ISSUE #1022: NSRDB advanced download sets time zone to 0 for full-disc and typical year
### Describe the bug

When downloading weather through the advanced download, typical year files (TGY, TMY, TDY) and fill-disc files have time zones of GMT 0 instead of the local time zone.

### To Reproduce

1. Go to the advanced download tool in the detailed PV model
2. Find a location
3. Select a full-disc weather file and/or a typical year weather file and download them
4. See that the resulting time zone is GMT 0

### Expected behavior

I expected the time zone to match the local time zone offset, as it does for other weather file types through the standard and advanced download tools.

 - OS: Windows 10
 - Version: SAM 2021.12.02 and 2021.12.02r1

### Additional context

- Downloading the default TMY file through the standard download tool (not the advanced one) works as expected. 
- It looks like the issue should have been fixed in [#609](https://github.com/NREL/SAM/pull/609)? 
- Maybe the issue for full-disc was fixed in [#950](https://github.com/NREL/SAM/pull/950)?

brtietz:
# Pull Request Template

## Description

Queries to psm3-tmy were coming back with no timezone information. When I tried the query in a browser, I got the following error:

image

Remove the problematic attrs, which now matches the lk script for the quick download.

Fixes #1022 

To test:

- Build SAM and SAM-private
- Download a psm3-tmy file
- The downloaded file has a timezone

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/nsrdb.cpp
+++ b/src/nsrdb.cpp
@@ -373,9 +373,9 @@ void NSRDBDialog::OnEvt( wxCommandEvent &e )
 			wxLogStatus("downloading (%d of %d): %s", (int)(i+1), (int)arychecked.Count(), (const char*)url.c_str());
 #endif
 							bool ok = curl.Get(url+attr);
-							// try without attributes
-							if (ok && (curl.GetDataAsString().Length() < 1000)) 
-								ok = curl.Get(url);
+							// in case attributes list is incorrect try without attributes if returned file is not a weather file
+							if (ok && (curl.GetDataAsString().Length() < 1000)) // file likely json or html instead of csv
+								ok = curl.Get(url+"&utc=false"); // without attributes returns file with all attributes
 							if (!ok)
 								wxMessageBox("Download failed.\n\n" + curstr +"\n\nThere may be a problem with your internet connection,\nor the NSRDB web service may be down.", "NSRDB Download Message", wxOK, this);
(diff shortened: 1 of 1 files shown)


cpaulgilman:
This fix is consistent with the NSRDB API documentation for the psm3-tmy endpoint showing that clearsky attributes are not available: https://developer.nrel.gov/docs/solar/nsrdb/psm3-tmy-download/

Other endpoints, such as psm3 have clearsky attributes that we want to include when available to address this SSC issue: https://github.com/NREL/ssc/issues/151

cpaulgilman:
When a call to the NSRDB fails, SAM tries a second call with no attributes in case the attribute list was incorrect. The reason the psm3-tmy call with incorrect attributes returned a file with UTC time zone is because the subsequent call with no attributes did not include the UTC parameter. That is fixed in https://github.com/NREL/SAM/commit/85930304705f7f3c0a8e185044b2a6a4af945bda.

cpaulgilman (approved):
This should be good after my commit adding the UTC attribute to the second API call.

Issue #836, opened 2021-12-02
PVWatts Merchant Plant Bugs


dguittet:
### Describe the bug

1. Running PVWatts Merchant Plant with default configuration results in missing values in cash flow

![image](https://user-images.githubusercontent.com/29387736/144452491-7c815de6-b9e6-4f33-a362-12cc546115a0.png)

2. Running with "Energy market cleared capacity and price" from simulation for 1 Year

    Steps: 1) Open new PVWatts Merchant Plant. Click Simulate. Copy "System power generation (KW)" from Time Series window. In Excel, divide that by 1000 to convert to MW. Paste MW values into "Energy market cleared capacity and price". Change project life on Financial Parameters page to 1 year. Click Simulate. It fails with this error:
    ```
    exec fail(merchant plant): The generation is not sufficient to meet the ancillary markets requirements.  
    Specifically, sum of cleared capacity 17.8484 MW does not match system capacity 17.8484 MW at timestep 8
    ```

3. Running with no Revenue selected results in hard crash

   Steps: Open new PVWatts Merchant Plant. Toggle "Enable energy market revenue" to disable. Click Simulate. Results in hard crash: 
    
    
![image](https://user-images.githubusercontent.com/29387736/144461859-1490835c-ac47-449a-a2ff-70d66f9c85fc.png)

   

### Platform

MacOSX, Open source build from 12/2/21

sjanzou:
Project file for testing with PVWatts annual output for year 1 cleared capacity 
[SAM_836.sam.zip](https://github.com/NREL/SAM/files/8644874/SAM_836.sam.zip)


See SAM pull request https://github.com/NREL/SAM/pull/1034 and project file for issue 1 

See ssc pull request https://github.com/NREL/ssc/pull/813 and project file for issues 2 and 3.

PR #1034, opened 2022-05-07
Addresses issue 1 in SAM #836 and goes with ssc branch SAM_836

sjanzou:


sjanzou:
Goes with ssc pull request 813 https://github.com/NREL/ssc/pull/813 and SAM_836 branches of ssc and SAM.

Issues are addressed with attached project file that has a case for each issue in https://github.com/NREL/SAM/issues/836 
[SAM_836.sam.zip](https://github.com/NREL/SAM/files/8644869/SAM_836.sam.zip)

sjanzou:
Note that original failing github actions was due to defaults results not updated. They are now updated and the ssc tests are being updated to give more reasonable LCOE results for merchant plants (shown on right in image below):
image

brtietz (approved):
SAM side of this fix looks good. Some cleanup on SSC required, see that PR.

cpaulgilman (approved):

PR #1043, opened 2022-05-17
along with ssc branch SAM_675, this fixes SAM issue 675

sjanzou:


brtietz (approved):
Fix makes sense - but why the big increases in LCOE in the test results?

cpaulgilman:
This is a bit of a nit-pick, and maybe not worth the time to fix, but would be good for the order of Mode options to be logical:

image

Perhaps:
 
```
Subhourly
Hourly
Daily
Weekly
Monthly
Annually
```

sjanzou:
> Fix makes sense - but why the big increases in LCOE in the test results?

@brtietz, running the patch 1 release and the SAM_675 branch shows that all the Windows defaults results match except PV Battery / SO which are close:
image


So, I assume that the previous test_results_win64.csv were somehow corrupted by a previous merge.

sjanzou:
> This is a bit of a nit-pick, and maybe not worth the time to fix, but would be good for the order of Mode options to be logical:
> 
> image
> 
> Perhaps:
> 
> ```
> Subhourly
> Hourly
> Daily
> Weekly
> Monthly
> Annual
> ```

@cpaulgilman, I have created a separate issue https://github.com/NREL/SAM/issues/1046 for rearranging as the original control had weekly and annual as optional mode selections and the enums for the list do not line up.

PR #1045, opened 2022-05-18
Change Metrics and Autograph labels to be more consistent with SSC labels

### LINKED ISSUE #969: Behind the meter summary page still uses "net" where it shouldn't
### Describe the bug

A number of metrics labels on the summary page still use the word "net energy to grid" for behind the meter systems, when they mean monthly AC energy:

image

image

### To Reproduce

1. Run a residential or commercial default case
2. Compare the summary metrics to the data tables

### Expected behavior

Labels should either match the data tables outputs, or eliminate the word "net"

 - OS: Windows
 - Version 2021.12.02 r1

cpaulgilman:
Fixes #969

"Annual AC Energy" for `annual_energy`

"Monthly AC Energy" for `monthly_energy'

Except SWH "Energy Savings" and IPH Models "Energy" (because thermal)

Also "Electricity Net Generation" for `cf_energy_net`


brtietz (approved):
Looks great, thanks for the fix!

Issue #1003, opened 2022-04-15
Creating new case from sorted parametric runs can use wrong inputs


williamhobbs:
### Describe the bug

When I perform a parametric analysis, then sort the results based on an input or output, and then right-click a run number and select "Create new case", the correct inputs are not always passed to the new case. It appears that the new case is create based on the _un-sorted_ inputs, which may not match the sorted inputs. 

### To Reproduce

1. Create a **PVWatts, No financial** model
2. Go to the Parametrics analysis page
3. Add _Azimuth (degrees)_ as an input
4. Add _Annual AC output (kWh)_ as an output
5. For run 2, change azimuth to 270 degrees
6. Click "Run simulations"
7. Click the _Annual AC output (kWh)_ header to sort ascending (the run with Azimuth of 270 should have moved from the 2nd row to the first row)
8. Right click the 1st run (with azimuth of 270) and select "Create new case"
9. In the new case, browse to azimuth in System Design and see that it is set to 180, not 270

### Expected behavior

Sorted inputs should carry over to new cases. 

### Operating System and Version

 - OS: Windows 10
 - Version: SAM 2021.12.2r1

PR #1055, opened 2022-05-28
Address SAM issue 1003

sjanzou:


sjanzou:
File to test based on steps to reproduce in https://github.com/NREL/SAM/issues/1003
[SAM_1003.zip](https://github.com/NREL/SAM/files/8790814/SAM_1003.zip)

brtietz (approved):
The test case was a little confusing, since the case named parametrics 1 had the wrong inputs - but after running the parametrics and re-exporting everything worked as expected.

cpaulgilman:
This works as expected after you run parametric simulations, and fixes the issue in SAM 2021.12.02.

If you sort the column, right-click the row header and create a new case before running the parametric simulation, the inputs in the new case may not match the inputs in the row of the table. Options to resolve:

1. Leave as-is and explain in documentation.
2. Export correct inputs to the new case after sorting even if there are no parametric results in table.
3. If there are no parametric results in table, automatically run parametric simulations on clicking **Create new case** in shortcut menu.
4. Disable short-cut menu when there are no parametric results.

PR #1058, opened 2022-05-31
Updated Windows defaults results based on ssc pull request 826

sjanzou:


brtietz:
Branch name has a typo, so GH Actions tests will fail since they aren't pulling the correct SSC branch. See new PR here: https://github.com/NREL/SAM/pull/1059

PR #1061, opened 2022-05-31
API updates from latest patch branches

sjanzou:


brtietz (approved):
Thanks for running the updates!

PR #1076, opened 2022-06-20
PV Uncertainty updates

sjanzou:
Please test and update tool tips as necessary
Goes with wex pull request 142
https://github.com/NREL/wex/pull/142

### DIFF
--- a/src/pvuncertainty.cpp
+++ b/src/pvuncertainty.cpp
@@ -143,9 +143,9 @@ PVUncertaintyForm::PVUncertaintyForm( wxWindow *parent, Case *cc )
 
 
 	m_layout = new wxSnapLayout(this, wxID_ANY);
-    
+    m_layout->SetSizeHints(1200, 1200);
     wxBoxSizer *sizer_plots = new wxBoxSizer( wxHORIZONTAL);
-    sizer_plots->Add(m_layout, 1, wxALL|wxEXPAND, 5 );
+    sizer_plots->Add(m_layout, 1, wxALL|wxEXPAND, 0 );
     
     wxBoxSizer *sizer_main = new wxBoxSizer( wxVERTICAL );
 	sizer_main->Add( sizer_top, 0, wxALL|wxEXPAND, 5 );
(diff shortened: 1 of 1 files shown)


mjprilliman (commented):
image
Is anyone else having trouble reading the distribution information?

image
Getting this gray box over the left side when a new case is building. Is that to update P50/P90 versus PV Uncertainty text?

-I'm not seeing the distribution selections persist on save, close, and re-open the case. 

-How will this be handled in version upgrade? After making a file off of PVUncertainty and develop branches that file has issues being opened in patch 2. 
image

sjanzou:
> image Is anyone else having trouble reading the distribution information?
> 
This issue is fixed in the latest commit for windows

> image Getting this gray box over the left side when a new case is building. Is that to update P50/P90 versus PV Uncertainty text?
> 
This is fixed and was due to an additional dialog being created with the PV Uncertainty form.

> -I'm not seeing the distribution selections persist on save, close, and re-open the case.

Can you provide and example? Every case I tested and saved and reopened had all the distribution, weather file folder and pvalues changes saved and shown when reopened.

> 
> -How will this be handled in version upgrade? After making a file off of PVUncertainty and develop branches that file has issues being opened in patch 2. image

The project file is not backwards compatible since a new data structure (PVUncertaintyData) is stored in the project file. Old projects files can be read into the new structure due to the versioning in the Case::Read method. Olde versions of SAM cannot open the newly saved file since the case persistence version is now 7 (see case.cpp line 522).

Thanks for the detailed feedback!

mjprilliman:
> `Can you provide and example? Every case I tested and saved and reopened had all the distribution, weather file folder and pvalues changes saved and shown when reopened.`<

After more testing, changing distributions and then hitting the 'Run PV Uncertainty simulations' button causes the distributions to persist and close and re-open. Not hitting the 'Run' button after changing the distributions causes the distributions to revert to the previously run values on close and re-open. I think this behavior probably makes sense for the tool and should be left as it is but maybe just something to point out in Help.

mjprilliman (approved):
Everything looks good after the most recent commits.

sjanzou:
From 6/21/2022 meeting:
1. Save distributions without running
2. Enable user CDF
3. do not clear graphs with p Value change
4. remove vertical scrollbars in plot sizer

PR #1084, opened 2022-07-05
Fix typo in detailed PV system design help file

kandersolar:
Hello dear SAM team, this fixes a small typo in the Help page for the detailed PV model system design, presumably a copy/paste oversight from the similar GCR text a few lines down.


cpaulgilman (approved):
@kanderso-nrel Thanks for fixing this typo.

We use [Help & Manual (H&M)](https://www.helpandmanual.com/) to generate these help HTML files, so these changes will be overwritten the next time we generate them. I fixed the typo in the source files, so your fix will be included in the next version. 

Unfortunately, the Help source files are in a proprietary format that requires the H&M software to edit. We hope to convert to another platform, perhaps using Markdown with [Jekyll](https://jekyllrb.com), but that will require effort to convert from H&M.

kanderso-nrel:
Got it, thanks @cpaulgilman!  I'll keep that in mind for the future.

PR #1096, opened 2022-07-15
Save defaults as JSON issue 1095

sjanzou:


sjanzou:
Note that conflicts have to be resolved manually since JSON code added and git does not know how to handle...

sjanzou:
This pull request will be reopened after the deleted JSON functions from Github merge and restored and tested.

sjanzou:
Update and remove .txt defaults and run and match all test results on Windows - no changes in test_results_win64.csv

cpaulgilman (approved):
Looks good! I tested the following:

* Changing default values in Defaults Manager
* Changing default values in text editor
* Ran test LK script to load_defaults() and simulate() for all configurations. No default messages except PTES, DSPT, and Dish Stirling which have "Defaults likely out of date" messages as expected because they are retired or not being maintained (PTES)

PR #1127, opened 2022-08-11
PV land area UI cleanup

### LINKED ISSUE #1125: PV module area estimate in hectares incorrect
### Describe the bug

The value in hectares on the Detailed PV model's System Design page is the total module area shown at the top of the page (single module area * number of modules) converted from m2 to ha. It should be the total module area estimate in acres converted to hectares.

Need to check this for both Detailed PV and PVWatts.

Also, "Total module area estimate" and "total module area" are different values. Need to improve the labels to make clear that the first is the module area projected onto the ground (having it under the "Land Area" group label is not enough to avoid confusion).

cpaulgilman:
Fix #1125 

* Fix mistake in calculation of array area in hectares (only affects UI calculated value).
* Add new UI calculated variables to show module area in acres and hectares. This makes it a little clearer how module area relates to total land area, and makes the effect of GCR and tilt angle more apparent when the user changes those inputs.
* Reorganize PV land area inputs and edit UI text.
* Regenerate API files for new UI inputs (new inputs should not affect SSC).
* Help revisions will be done separately.

### DIFF
--- a/api/api_autogen/library/defaults/Battery_FuelCellCommercial.json
+++ b/api/api_autogen/library/defaults/Battery_FuelCellCommercial.json
@@ -163,6 +163,8 @@
 		},
 		"SystemCosts": {
 			"om_batt_replacement_cost": [0.000000],
+			"om_batt_variable_cost": [0.000000],
+			"om_production_escal": 0,
 			"om_replacement_cost_escal": 0
 		},
 		"FuelCell": {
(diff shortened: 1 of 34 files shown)


mjprilliman (approved):
Github seems to be having some trouble loading the diffs for the UI forms and api module files but the changes look good to me on the build. 

image

cpaulgilman:
The diffs loaded ok for me:
image

PR #1128, opened 2022-08-12
General test generation framework lk function simulate_ssc_tests

sjanzou:
Goes with https://github.com/NREL/ssc/pull/858. lk function fcall_simulate_ssc_tests in script.cpp generates two json files from SAM case runs. The first file contains all inputs to run a single compute module independently of other compute modules. The second json file contains all outputs from the specified compute module to allow for ssc test generation.

### DIFF
--- a/deploy/generate_test_json.lk
+++ b/deploy/generate_test_json.lk
@@ -1,0 +1,102 @@
+// naming convention with date and github branch used
+// configuration and compute module appended in simulate_ssc_tests
+casename =  "2022.08.08_develop_branch";
+
+// update folder based on where you would like the generated json files (one for input and one for output comparison for each configuration)
+// macOS
+folder = "/Users/imacuser/Public/Projects/SAM/Documentation/FinancialMarkets/Tests/generated2";
+// Windows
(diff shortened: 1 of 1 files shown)


sjanzou:
Goes with https://github.com/NREL/ssc/pull/858 

The attached lk file  can be used with a SAM project file to generate json files to use with a general googletest framework for comparison number and array outputs. See cmod_cashloan_test.cpp for an example use of the generated json files.
[tests.zip](https://github.com/NREL/SAM/files/9315637/tests.zip)
.

brtietz:
Can tests.lk be committed to the branch? (any reason not to?) I'm not sure which folder, though intuitively I'd expect it in the same place as test_script_ow.lk. Maybe generate_test_json.lk is a better name for the new script?

sjanzou:
> Can tests.lk be committed to the branch? (any reason not to?) I'm not sure which folder, though intuitively I'd expect it in the same place as test_script_ow.lk. Maybe generate_test_json.lk is a better name for the new script?

Added generate_test_json.lk script to SAM/deploy folder as requested.

PR #1133, opened 2022-08-19
Battery power target dview, decimal fix for reopt sizing

mjprilliman:
## Description

-Added Dview viewer for battery power targets. Identified during SAM Open House demonstration. 
-Set decimal limit of 3 for battery power sizing for ReOpt battery sizing. 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

If you have added a new compute module in a SSC pull request related to this one, be sure to check the [Process Requirements](https://github.com/NREL/SAM/wiki/Compute-modules-in-SAM).

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1134, opened 2022-08-20
Fix build issue on macOS 12.5 on M1 with CMakeList update for 1129

sjanzou:


sjanzou:
Also fixed in pull request #1133

dguittet (approved):
LGTM

PR #1176, opened 2022-10-07
Fix SAM issue 1175 in variables.cpp - other changes in export_config and SAM_api

sjanzou:


brtietz (approved):
Thanks for the fix and the export config & API updates!

PR #1179, opened 2022-10-08
Revert SingleThreadHandler::WriteDebugFile to patch version to suppress prompting for lk files

sjanzou:


sjanzou:
Now SingleThreadHandler (used in fcall_reopt_size_battery) does not prompt for lk debug script generation. 
New class SingleThreadHandlerWithDebugOutput prompts for debug output and can be used if desired.
.Functionality not works as previous release.

cpaulgilman (approved):
This fixes the issue in Develop of the file prompts when running REopt.

brtietz (approved):
Thanks for the fix!

Issue #1072, opened 2022-06-07
Rankine model performance is insensitive to boiler pressure


qualand:
### Problem:

Results from SAM models using the Rankine cycle are insensitive to the user input of boiler pressure. This behavior is by design because SAM has no internal relationships between cycle boiler pressure, cycle design performance, and cycle cost. SAM relies on users to update these values base on the system they are trying to model.

### Solution:

In the current state, this model behavior is not clear to the user. Specifically, the purpose and/or behavior of boiler pressure specification is not well documented. This issue could be handle by either of the following methods:
1. Update documentation to clearly state the purpose of the boiler pressure input and that the user should update design efficiency and power cycle cost assumptions.
2. Remove boiler pressure for the SAM GUI, provide models with a nominal boiler pressure, and update documentation. 

Both of these solutions would require a characterization of the purpose of boiler pressure within the current SAM Rankine cycle model.  

### Quick Parametric Results (CSP Tower Molten Salt - Rankine Cycle)

LCOE (real):
![image](https://user-images.githubusercontent.com/34353104/172441577-f9f37ff6-91d9-4489-9298-b979f87ac5ef.png)

Capacity Factor:
![image](https://user-images.githubusercontent.com/34353104/172441688-f686df93-dc8b-4f28-9992-c7cdbcfab036.png)

Note: The versions of the Rankine cycle model are used across multiple SAM models.

cpaulgilman:
See https://sam.nrel.gov/component/kunena/forum-general/3725.

If we go with Option 1, it would be good to have some guidance on the how to change the design efficiency, perhaps a table of suggested values or a specific example.

PR #1192, opened 2022-10-14
Remove boiler pressure as input from indirect cycle model

tyneises:
Corresponds to issue #1072

We removed boiler pressure (P_boil) as a user input to the indirect Rankine Cycle model, which is used in the molten salt power tower, physical trough, molten salt fresnel, and ETES models. Because the cycle efficiency and HTF outlet temperatures are provided as an independent inputs by the user, the boiler pressure input does not modify these values (as one might expect as the first-order effect of boiler pressure). Instead the model uses boiler pressure in second order calculations to 1) define a boiling temperature to normalize off-design HTF temperature and 2) estimate steam mass flow for cycle make-up water calculations. Because boiler pressure only influences results in these minor non-intuitive ways, we decided to hardcode the value to 100 bar.

We do not expect this change to significantly influence results for saved cases of the relevant technologies, with the possible exception of trough cases where the user changed the boiler pressure to a value larger than the default value of 100 bar. In that scenario, this change will likely increase the annual energy of the system due to improved off-design power cycle efficiency.


tyneises:
Also deleted some unused UI pages and removed DSG references in the solar field UI page callbacks

PR #1206, opened 2022-10-24
Updates UI and defaults for ITC array inputs

sjanzou:
Goes with ssc pull request 901

### DIFF
--- a/deploy/runtime/ui/Financial Tax Credits.txt
+++ b/deploy/runtime/ui/Financial Tax Credits.txt
@@ -663,8 +663,8 @@ Tool Tip
 0
 Label
 5
-5
-Value
+6
+$/year
 UseSchedule
 2
 0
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
UI implementation looks good. I fixed widget alignment, revised schednumeric descriptions, and set table length to analysis period for all tax credit widgets.

brtietz (commented):
Do we want to change any percentages here? Wind is now receiving both the PTC and ITC, which is not correct. Battery is receiving an ITC, which is fine under the IRA, but it should be 30%.

My intuition is it'd be better to keep the numbers the name in this PR, and then have another dedicated PR for defaults updates.

sjanzou:
> Do we want to change any percentages here? Wind is now receiving both the PTC and ITC, which is not correct. Battery is receiving an ITC, which is fine under the IRA, but it should be 30%.
> 
> My intuition is it'd be better to keep the numbers the name in this PR, and then have another dedicated PR for defaults updates.

Separate pull request is my vote

PR #1209, opened 2022-10-25
SAM issue 1194 - api and test results changes (messages)

sjanzou:
Goes with ssc pull request 903


cpaulgilman (approved):
Looks good. I built PySAM and PySAM documentation to check ITC as array is working in Python script and correctly documented as a sequence in documentation.

PR #1211, opened 2022-10-26
Update bifaciality defaults based on market increases

brtietz:
Get this update done early for PVWatts release.


cpaulgilman (approved):
Thanks for making this change.

PR #1227, opened 2022-11-02
Read ITC array values in output reports

### LINKED ISSUE #1226: Output reports printing nan ITC
### Describe the bug

ITC federal and state variables being changed to arrays breaks the report templates and prints nans. I am working on a fix. 

### To Reproduce

1. Go to Detailed PV, Single Owner
2. Create Report
3. Look at incentives box on Page 1

### Expected behavior

Reports should work similar to PTC printing in reports

mjprilliman:
# Pull Request Template

## Description

-Change report templates to read ITC federal and state as arrays rather than numbers. 

Fixes #1226 

## Type of change

## Checklist:

### DIFF
--- a/test_results_win64.csv
+++ b/test_results_win64.csv
@@ -61,7 +61,7 @@ Standalone Battery,Leveraged Partnership Flip,-8397703,79.0709,88.6283,; Warning
 Standalone Battery,All Equity Partnership Flip,-8397703,82.9781,88.6283,; Warning: IRR at end of analysis period is not a number (NaN). This can indicate that revenues are too low to cover costs, or that they are excessively high compared to costs.
 Standalone Battery,Sale Leaseback,-8397703,83.5501,88.6283,; Warning: IRR at end of analysis period is not a number (NaN). This can indicate that revenues are too low to cover costs, or that they are excessively high compared to costs. Warning: NPV is $-1.09966e+08. A negative NPV indicates project costs are higher than revenues.
 ETES,Single Owner,-497823968,15.5498,12.4517,; 
-PTES,Single Owner,-203106896,17.4526,12.9877,;  
+PTES,Single Owner,-203106896,17.4526,12.9877,; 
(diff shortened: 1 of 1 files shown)


cpaulgilman (changes requested):
When ITC is an array of more than one value, I don't think it makes sense to report the average over the analysis period in the PDF report. (The PBI applies to annual production every year, so reporting the average over the analysis period is somewhat meaningful, but the ITC applies to the installation cost only in the years where the percentage is not zero.) For ITC it might make more sense to report the sum of the array values instead of their average, and put "total" in parentheses instead of "average".

@sjanzou, what do you think?

mjprilliman:
Updated mini-grid example to latest version as otherwise Version upgrade window is blocking user interface and can't close it without exiting the wizard first. Thoughts @sjanzou @cpaulgilman ?

cpaulgilman:
> Updated mini-grid example to latest version as otherwise Version upgrade window is blocking user interface and can't close it without exiting the wizard first.

I agree with updating the mini-grid-example.sam file to the current version so the upgrade script does not run -- we should add that to the pre-release checklist. I think that is the only .sam file in the distribution.

sjanzou:
> Updated mini-grid example to latest version as otherwise Version upgrade window is blocking user interface and can't close it without exiting the wizard first. Thoughts @sjanzou @cpaulgilman ?

Thanks for updating the mingrid sam file - that should be one more thins off of the pre-release checklist!

sjanzou:
> When ITC is an array of more than one value, I don't think it makes sense to report the average over the analysis period in the PDF report. (The PBI applies to annual production every year, so reporting the average over the analysis period is somewhat meaningful, but the ITC applies to the installation cost only in the years where the percentage is not zero.) For ITC it might make more sense to report the sum of the array values instead of their average, and put "total" in parentheses instead of "average".
> 
> @sjanzou, what do you think?

@cpaulgilman, yes, I agree, the sum is the most meaningful single metric for the ITC array.

PR #1245, opened 2022-11-10
Atb defaults update script

brtietz:
Python script used to update the defaults using the file at https://github.com/NREL/SAM-documentation/tree/master/Defaults%20References%20and%20Tools/Fall%202022

Given our decision making on defaults (PV benchmark for PV, ATB on others) there will be some further modifications to the scripts throughout the process.

Defaults work in progress.

### DIFF
--- a/samples/ATB Defaults Update/updaters.py
+++ b/samples/ATB Defaults Update/updaters.py
@@ -81,9 +81,9 @@ class WindUpdater(BaseUpdater):
         super().__init__(tech_name, 2020)
 
     def update_defaults(self, defaults_json, atb_df):
-        defaults_json = super().update_defaults(defaults_json, atb_df)
+        defaults_json, atb_df = super().update_defaults(defaults_json, atb_df)
         # Extract additional parameter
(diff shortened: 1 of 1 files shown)


cpaulgilman (commented):
Looks good. The script runs and sets defaults for Detailed PV and PVWatts configurations.

1. One small thing in the readme: The script requires pandas and python-rapidjson.

2. I'm having trouble matching up values I see in SAM with the script output and ATB spreadsheet. The utility costs seem to match, but the residential don't. Not sure if I'm looking in the right places:

image'

image

3. It might be useful to append a column to the CSV input file with the defaults generated by the script so we have a record, and can easily compare SAM defaults with the source data.

brtietz:
>     2. I'm having trouble matching up values I see in SAM with the script output and ATB spreadsheet. The utility costs seem to match, but the residential don't. Not sure if I'm looking in the right places:

The issue there is that the version of the CSV in this PR includes "incl. supply chain premium" which causes the script to skip the value. I fixed that in the csv in https://github.com/NREL/SAM/pull/1248

I'll get the other changes pushed here.

Issue #1251, opened 2022-11-11
ETES version upgrade equations do not evaluate


sjanzou:
Test file to test 
[2021_SAM_Defaults.zip](https://github.com/NREL/SAM/files/9989207/2021_SAM_Defaults.zip)

To reproduce
1. Open attached project file in SAM 2022.11.21 release build.
<img width="1220" alt="image" src="https://user-images.githubusercontent.com/6498311/201317890-54a212e6-3ea1-4544-976e-c3582efab59a.png">

<img width="392" alt="image" src="https://user-images.githubusercontent.com/6498311/201318213-d58bb9f9-266f-4b32-b67e-2f54624ea1b1.png">

PR #1256, opened 2022-11-13
Fix SAM issue #1251

sjanzou:


tyneises (approved):

tyneises:
@sjanzou I removed the line in the ETES cmod mimo that sets a local weather file to a ssc_var. It was commented out. Please revert if necessary.

PR #1257, opened 2022-11-15
UI layout fixes Part 1

cpaulgilman:
UI layout fixes for SAM on Ubuntu. Need to go through more UI pages, but get this merged now to minimize potential merge conflicts.

- Location and Resource
- New PV albedo options
- CEC module with database
- CEC module with user specs
- Inverter with database
- Inverter with user specs
- PV system design
- PV shading and layout
- PPA revenue capacity payments


sjanzou (changes requested):
Inverter Temperature Derate Curves table is curtoff on macOS 13.0
image

CEC User module page I-V curve legend at bottom goes off end
image

All other changes look great on macOS 13.0!

sjanzou (approved):

sjanzou:
Approved changes so as not to block merging and the two requested form changes above.

cpaulgilman:
@mjprilliman There is a "Inverter Temp Derate" UI form that is not being used. Instead there is "Inverter Temp Derate CEC DB" and other copies of the form to handle different variable names (`inv_tdc_cec_db`, `inv_tdc_cg`, etc.) and sizes of widgets. Is it OK if I delete that original "Inverter Temp Derate" form with its unused `inv_tdc_data`?

mjprilliman:
> @mjprilliman There is a "Inverter Temp Derate" UI form that is not being used. Instead there is "Inverter Temp Derate CEC DB" and other copies of the form to handle different variable names (`inv_tdc_cec_db`, `inv_tdc_cg`, etc.) and sizes of widgets. Is it OK if I delete that original "Inverter Temp Derate" form with its unused `inv_tdc_data`?

@cpaulgilman Sure I see no reason to keep the unused 'Inverter Temp Derate' so I think it is fine to delete it

leo-smi:
My layout are messed up in linux deepin 20
sam: 2022.1.21

the system font chaging causes this. I think SAM can use independent font size from the system to avoid this "issue".

befora change to minimum font sizer

Capturar_20230218204050

after changing to minimum font size

image

PR #1248, opened 2022-11-11
Fall 2022 defaults updates: part 1

brtietz:
## Description

Defaults update for Fall 2022. Update includes:

* Updated PV, Battery, and Wind CAPEX, O&M numbers, and system sizes
* Updated incentives for all technologies based on IRA "base case"

This is part 1.

Once we approve these costs, I can update revenue to correct issues with IRR or NPV as possible
(ex: cases that didn't have negative NPV but now do due to increased costs). (target 11/14)

Part 3 will include updating the calculated values in the defaults and export for PySAM (target 11/15)

References available in: https://github.com/NREL/SAM-documentation/tree/master/Defaults%20References%20and%20Tools/Fall%202022

When a $/kW value was available, it was used with inflation to 2022 dollars

When only a $ value was available (often the case for batteries) I attempted to maintain fidelity with the ATB spreadsheet
projections first, and the benchmark second, targeting the total dollar amount in the benchmark.

Issue #955, opened 2022-02-24
Add Capacity Payment Defaults to More Technologies?


janinefreeman:
We should consider updating the capacity payment defaults from zero for a variety of technologies, based on current industry trends. See spreadsheet in sam-documentation/Defaults References and Tools/Capacity_value_methods.xlsx for additional thoughts.

cpaulgilman:
See also https://github.com/NREL/SAM/issues/955

brtietz:
I would love to use an NREL source for this, but the Cambium data defines the capacity costs at a system level, not at a plant level: https://www.nrel.gov/docs/fy22osti/81611.pdf, and the spreadsheet in developer tools includes a lot of references that predate the deployment of Li-Ion batteries for this purpose.

This EIA source seems reasonable, and would suggest a value of $54,000 /MW-yr for batteries: https://www.eia.gov/outlooks/aeo/IIF_battery/

Thoughts on whether this year's update should be for just batteries or more technologies?

brtietz:
https://github.com/NREL/SAM/pull/1258 added capacity payment defaults to relevant stand alone batteries. Closing this issue for this release. We should continue to review capacity payment defaults along with other revenue defaults, especially after https://github.com/NREL/SAM/issues/635 is completed.

PR #1258, opened 2022-11-15
Fall 2022 defaults part 2: solve financial issues

brtietz:
Update revenue inputs and solve a few other issues after https://github.com/NREL/SAM/pull/1248

- Detailed PV - None now matches single owner size, inverter, and module
- Host developer PPA price and utility rates are now consistent within pv-only and battery
- Added capacity payments for stand alone battery single owner and merchant plant (see https://github.com/NREL/SAM/issues/955)
- Additional PPA price adjustments to improve NPV

Overall the approach leaned on price consistency within cases such as "pv battery - front of meter." If something was not profitable in 2021.12.02, I didn't put in a lot of effort to make it profitable here.

I recommend a lot of attention to the changes in warnings in test_results_win64.csv when reviewing


mjprilliman (approved):
-Detailed PV - Host Developer is in "Specify PPA Price" mode, while PVWatts - Host Developer is in "Specify IRR Target" mode. It may be fine to leave that way but just a note. Might be worth double checking ppa_soln_mode is set correctly on PV configs.

mjprilliman:
Standalone capacity payments right now applied to 0% of the battery nameplate by default.

mjprilliman:
> Standalone capacity payments right now applied to 0% of the battery nameplate by default.

-From Paul: make sure collapsible variable for UI forms is set to 1 for defaults with capacity payments

cpaulgilman (approved):

PR #1262, opened 2022-11-16
Fall 2022 defaults pt 3 update calculated values

brtietz:
Run script to overwrite defaults with latest calculated values for PySAM

Run export config and SAM api export for PySAM release

Target branch will be switched to develop after https://github.com/NREL/SAM/pull/1258 passes test and is merged


tyneises (approved):

mjprilliman (approved):
The LK script to do this is a great addition!

PR #1264, opened 2022-11-16
add script to calculate block schedules from 8760 pricing

tyneises:


mjprilliman (approved):
Worked with both csv files included.

cpaulgilman (approved):

PR #1266, opened 2022-11-16
Update wave download function to correctly handle cancel button press

mjprilliman:
# Pull Request Template

## Description
-Cancel button was causing callback error on wave download window as callbacks are looking for hash table. Corrected so that canceling creates hash table with empty quotes then exits. 

## Type of change


mjprilliman:
To test: 
-Open Marine Wave - Single Owner case
-On Wave Resource page, click 'Download and add to library...' button
-In window, click Cancel
-See no callback error

sjanzou (approved):

PR #1269, opened 2022-11-17
CEC libraries fall 2022

cpaulgilman:
Update CEC module and inverter libraries.

* I ran test_script_ow.lk, and updated modules and inverters do not change results for any PV configurations.
* Source data files and test results are in DocHub.

@sjanzou Feel free to let @mjprilliman review this if you've got other tasks.


mjprilliman (approved):
Libraries loaded in fine. Looks good on a quick scan in Excel. Running the LK test scripts locally now.

sjanzou (approved):
Thanks for updating and weeding out the bad apples!

PR #1271, opened 2022-11-17
Fix broken CMakeList.txt for MacOS build

sjanzou:


mjprilliman (approved):

brtietz (approved):
Thanks for catching this!

I re-ran the test script and the results for csp generic system didn't change. This change shouldn't affect that, though.

PR #1274, opened 2022-11-17
UI fixes for PVWatts

cpaulgilman:
* Fix "Weather file has snow depth data" message when weather file does not have snow data.

* Fix land area UI AC nameplate capacity value in kW instead of MW (land area calc, correct, just UI display value was incorrect)

* Fix UIOjbect property for advanced PVWatts inputs and widgets.


mjprilliman (approved):
Looks good. Thanks

PR #1279, opened 2022-11-18
Remove .txt default files

cpaulgilman:
Default .txt files were possibly unintentionally reintroduced by a merge removed. All default files should be .json files.


sjanzou (approved):
@brtietz,  We need to have everyone do a clean checkout of SAM so we do not get any .txt defaults checked in with pull requests in the future.

Issue #1285, opened 2022-11-18
Using logmsg in equations can cause SAM to crash


brtietz:
### Describe the bug

If uncommented, executing the code below can cause SAM to crash:

![image](https://user-images.githubusercontent.com/5530592/202812045-c839e9cd-b444-49d2-871a-d469f45935a0.png)

(from Battery Dispatch Options BTM)

### To Reproduce

1. Open a BTM PV-Battery Case
2. Select "peak shaving dispatch" with a custom load profile and custom load escalation
3. Switch to another dispatch option
4. SAM crashes

### Expected behavior

SAM should not crash

Workaround for the release is to comment out usages of logmsg in equations. We should discuss whether or not to allow this in lk callbacks in the future.

### Operating System and Version

 - OS: 2021.12.02 and 2022.11.21 RC2

brtietz:
Closed in #1286

PR #1286, opened 2022-11-18
Comment out logmsg as quick fix

brtietz:
The logmsg in the battery equation was causing SAM to crash https://github.com/NREL/SAM/issues/1285

Should this be a msgbox, or something else?

This was the only instance of this in the UI.


cpaulgilman (approved):
That looks like a left-over debugging line for development. I don't think it needs to be a message box.

PR #1292, opened 2022-12-02
change variables type to calculated

tyneises:
A few UI variables in the MSPT model were not correctly labeled as calculated

And delete unused variable


brtietz:
Same comment as https://github.com/NREL/ssc/pull/949 - develop, not patch?

tyneises:
> Same comment as [NREL/ssc#949](https://github.com/NREL/ssc/pull/949) - develop, not patch?

Thanks, develop (not patch) is fine

PR #1302, opened 2022-12-10
Fix run_defaults_diff

dguittet:
## Description

run_defaults_diff.py needed to be updated to handle empty strings

## Type of change


brtietz (approved):
Thanks for the fix!

PR #1331, opened 2023-01-26
Openei eulp take 2

brtietz:
https://github.com/NREL/SAM/pull/1300, but with a clean commit history


cpaulgilman (approved):
Sample script works -- looks like everything got cherry-picked correctly. Thanks!

PR #1338, opened 2023-01-30
UI forms as JSON

sjanzou:
goes with associated pull requests in wex 155 and SAM-private

### DIFF
--- a/src/ide.cpp
+++ b/src/ide.cpp
@@ -831,9 +831,9 @@ UIEditorPanel::UIEditorPanel( wxWindow *parent )
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_TEXT, "Load text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_ALL_TEXT, "Save all text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 	//sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_ALL_TEXT, "Load all text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
-	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_JSON, "Save JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
+//	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_JSON, "Save JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_JSON, "Load JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
-	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_ALL_JSON, "Save all JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
+//	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_ALL_JSON, "Save all JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_ALL_JSON, "Load all JSON", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 	sz_form_tools->AddStretchSpacer();
 	sz_form_tools->Add( new wxButton( this, ID_VAR_REMAP, "Remap", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL|wxEXPAND, 2 );


mjprilliman (commented):
I like the json formatting of the UI files. How will we diff callback and equation changes in the future? Seems like it might be more difficult as a single string with \n values.

brtietz:
> I like the json formatting of the UI files. How will we diff callback and equation changes in the future? Seems like it might be more difficult as a single string with \n values.

That's a great point. The best recommendation for this seems to be using an array of strings, in which each element of the array is a line: https://stackoverflow.com/a/52557585

If we want to go that route I would recommend completing the review of this format and migrating to that format as a separate issue, so we're a little less locked down on forms changes for the patch.

src/variables.cpp
```diff
@@ -2348,7 +2347,8 @@ void VarInfo::Write(wxOutputStream &os)
 	wxDataOutputStream out(os);
 	out.Write8(0xe1);
 	//	out.Write8(2);
-	out.Write8(3); // change to version 3 after wxString "UIObject" field added
+//	out.Write8(3); // change to version 3 after wxString "UIObject" field added
+	out.Write8(4); // change to version 3 after wxString "UIObject" field added
```
brtietz, line 2351: Comment doesn't make sense here - what is version 4?

src/variables.cpp
```diff
@@ -2482,9 +2515,112 @@ bool VarInfo::Read_text(wxInputStream &is)
...
+	// Type
+	Write_JSON_value(doc, "Type", Type);
+	// Label
+	Write_JSON_value(doc, "Label", Label);
+	// Units
+	Write_JSON_value(doc, "Units", Units);
+	// Group
+	Write_JSON_value(doc, "Group", Group);
+	// IndexLabels
+	/* Handle multiline equations in IndexLabels
```
brtietz, line 2550: Nice comment!

src/ide.cpp
```diff
@@ -801,17 +817,24 @@ UIEditorPanel::UIEditorPanel( wxWindow *parent )
...
+	//	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_ALL, "Save all", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_ALL, "Load all", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_TEXT, "Save text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_TEXT, "Load text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
 //	sz_form_tools->Add(new wxButton(this, ID_FORM_SAVE_ALL_TEXT, "Save all text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
-//	sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_ALL_TEXT, "Load all text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
+	//sz_form_tools->Add(new wxButton(this, ID_FORM_LOAD_ALL_TEXT, "Load all text", wxDefaultPosition, wxDefaultSize, wxBU_EXACTFIT), 0, wxALL | wxEXPAND, 2);
```
brtietz, line 833: Maybe I'm missing something, but why add commented out lines? This is hard to follow with all the commented out code.

brtietz (changes requested):
This is really exciting, but I think it needs a little more polish prior to merging. For example: why are there both "Save" and "Save as JSON" buttons in the IDE?

image

I feel like there should only be one save button at merge time. Or is there a reason we'll need both for a while?

brtietz:
> > This is really exciting, but I think it needs a little more polish prior to merging. For example: why are there both "Save" and "Save as JSON" buttons in the IDE?
> > image
> > I feel like there should only be one save button at merge time. Or is there a reason we'll need both for a while?
> 
> Why am I not seeing these additional buttons when I build the branches?



> > This is really exciting, but I think it needs a little more polish prior to merging. For example: why are there both "Save" and "Save as JSON" buttons in the IDE?
> > image
> > I feel like there should only be one save button at merge time. Or is there a reason we'll need both for a while?
> 
> Why am I not seeing these additional buttons when I build the branches?

Oh, whoops. I had an older version of SAM 504 checked out.

mjprilliman:
Clicking out of a UI form in the IDE registers a diff in the json without hitting save or making any changes. Diff software cannot find any differences in the files but the file is shown to be modified. 

image

cpaulgilman:
This looks good from the UI. 

I also tested loading all configs with this LK script: 

```
techs = list_technologies();
for (t=0;t<#techs;t++)
{
	fins = list_financing(techs[t]);
	for (f=0;f<#fins;f++)
	{
		outln(techs[t] + ', ' + fins[f]);
		configuration(techs[t],fins[f]);
	}	
}
```

The script works in Windows, but on Linux (Mint 21.1), when I run the script, I get errors like:

* Solar Resource Data: "could not find library:SolarResourceData" when loading the Solar Resource Data
* Marine Wave Resource Time Series: "Error: 'PacWave East' is not available in the library. Choose a different item."
* Wind Resource File: "Error: 'WY Cheyenne_2006' is not available in the library. Choose a different item."

But the forms load fine when opening a case by hand, so not sure it's worth pursuing at this stage.

src/ide.cpp
sjanzou: > Maybe I'm missing something, but why add commented out lines? This is hard to follow with all the commented out code.

The code is temporarily commented so that when merging into develop, the updated text forms can be loaded and then saved as JSON and then the text UI forms can be removed. The forms in develop have diverged from patch so that a separate pull request is necessary for the SAM repo.

sjanzou:
> Clicking out of a UI form in the IDE registers a diff in the json without hitting save or making any changes. Diff software cannot find any differences in the files but the file is shown to be modified.
> 
> image



> Clicking out of a UI form in the IDE registers a diff in the json without hitting save or making any changes. Diff software cannot find any differences in the files but the file is shown to be modified.
> 
> image

Maybe we can write a better code versioning system than github that sometimes defaults to date and time changes instead of actually comparing file contents...

sjanzou:
> This looks good from the UI.
> 
> I also tested loading all configs with this LK script:
> 
> ```
> techs = list_technologies();
> for (t=0;t<#techs;t++)
> {
> 	fins = list_financing(techs[t]);
> 	for (f=0;f<#fins;f++)
> 	{
> 		outln(techs[t] + ', ' + fins[f]);
> 		configuration(techs[t],fins[f]);
> 	}	
> }
> ```
> 
> The script works in Windows, but on Linux (Mint 21.1), when I run the script, I get errors like:
> 
> * Solar Resource Data: "could not find library:SolarResourceData" when loading the Solar Resource Data
> * Marine Wave Resource Time Series: "Error: 'PacWave East' is not available in the library. Choose a different item."
> * Wind Resource File: "Error: 'WY Cheyenne_2006' is not available in the library. Choose a different item."
> 
> But the forms load fine when opening a case by hand, so not sure it's worth pursuing at this stage.

Where are you running SAM from to get this script error? Did you make from the SAM-private build_linux folder and run from the installation folder? I am not seeing this issue on MacOS or on CentOS7.

.github/workflows/ci.yml
```diff
@@ -207,11 +211,11 @@ jobs:
       run: |
         echo "GITHUB_REPOSITORY_OWNER=${GITHUB_REPOSITORY_OWNER}" >> $GITHUB_ENV
         git ls-remote --heads --exit-code https://github.com/${GITHUB_REPOSITORY_OWNER}/lk.git $GIT_BRANCH
-        if [[ $? != "0" ]]; then echo "LK_BRANCH=patch" >> $GITHUB_ENV; else echo "LK_BRANCH=$GIT_BRANCH" >> $GITHUB_ENV; fi
+        if [[ $? != "0" ]]; then echo "LK_BRANCH=develop" >> $GITHUB_ENV; else echo "LK_BRANCH=$GIT_BRANCH" >> $GITHUB_ENV; fi
```
brtietz, line 214: Please correct the CI target branch back to patch for this PR. We might need another dedicated change to make sure develop sticks on develop.

brtietz (changes requested):
It looks like the updates to the CI script swapped the target branch from patch to develop, that need to be addressed prior to merge.

Also the comment on line 2351 of variables.cpp is referring to the wrong version number

brtietz (approved):

brtietz:
The diff for changing lines looks great. The diff for adding a new line, however, is quite messy:

image

Is there a reason it has to be an object with line numbers as opposed to an array?

sjanzou:
Arrays should work now with the two str replace…. Both objects and arrays
had the same escaping sequence issue.

I can make the change and test new lines and update the pull request.

sjanzou:
After updating to RepidJSON array of strings, single line changes and new line added changes in AA Widgets results in desired diff viewing:
image

brtietz (approved):
This is great! One additional case not shown in Steve's tests, removal:

image

All of these diffs are perfectly readable, in my opinion.

PR #1352, opened 2023-02-11
PV Design UI Improvements

cpaulgilman:
# Pull Request Template

## Description

For configurations with Detailed PV model, make inverter voltage ratings editable on System Design page so users can correct voltage ratings from CEC databsae that don't match manufacturer datasheets.

Improve PV sizing messages with more detailed suggestions:

image

Make multiple MPPT option more transparent:

image

**To test:**

1. Create any Detailed PV or PV Battery case.
2. On system design page, change inverter voltage ratings or number of modules per string, and note improved sizing messages. Test with multiple subarrays.
3. Try changing inverter on Inverter page for different inverter models.
4. Verify MPPT voltage clipping in results, note simulation message if array voltage exceeds inverter Vdc_max.
5. Test with multiple MPPTs.

Fixes #1020

Partially addresses #209

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update
- [ ] This change modifies variables in existing compute modules. Please see [Checking for PySAM Incompatible API Changes](https://github.com/NREL/SAM/wiki/Checking-for-PySAM-Incompatible-API-Changes).

## Checklist:

If you have added a new compute module in a SSC pull request related to this one, be sure to check the [Process Requirements](https://github.com/NREL/SAM/wiki/Compute-modules-in-SAM).

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [X] I have checked my code and corrected any misspellings

Issue #28, opened 2017-10-26
Request up to 36 TOU periods for electricity rate structures


cpaulgilman:
This is for a CFE (Mexico) rate structure:

https://sam.nrel.gov/node/75299, now at https://sam.nrel.gov/forum/forum-general/1856

Note that the rates also change every year, so even with the additional TOU periods, the rate structure data will not match the actual rates.

brtietz:
Link to post on current forum https://sam.nrel.gov/forum/forum-general/1856-comision-federal-de-electricidad-tarifa-h-m.html#5794

It looks like the time series rates got implemented, does that cover this case?

cpaulgilman:
The time series energy buy and sell rates partially addresses this issue, but it does not address rates varying by year, which is another request we get periodically, such as in the following:

https://sam.nrel.gov/forum/forum-general/2891

https://sam.nrel.gov/forum/forum-general/2814

I'm not sure if it makes sense to add that level of complexity to the bill calculator. If we decide not to pursue that, we can close this issue.

Thanks,
Paul.

brtietz:
Adding to patch 2 based on a dGen request for Colombian rates where there is a weekday and a weekend period that varies every month of the year. Initial scope is to add this to SSC for PySAM, we may create a separate issue to add this to the GUI (need to decide if the GUI work is patch or release).

cpaulgilman:
Based on SSC implementation in https://github.com/NREL/ssc/pull/781, need to fix width of first column in Electricity Rate Data by Tier and Period table to fit 2-digit numbers. I would recommend making it adjustable like the other columns in the table:

![image](https://user-images.githubusercontent.com/27711933/158645585-ac777989-1e54-4442-a60a-8fc215638700.png)

![image](https://user-images.githubusercontent.com/27711933/158645736-5d875f24-9bad-4c5f-83af-8522a11d4919.png)

brtietz:
781 has been merged into SSC - leaving this issue open and moving to the fall release so we can discuss implications on the GUI. Areas to test/discuss:

- Copying to/from TOD/Battery Dispatch schedules
- Output sizing as Paul discusses above
- Others?

cpaulgilman:
SAM 2021.11.21 crashes when opening .sam file with 24 rates in https://github.com/NREL/ssc/files/8255563/24_rates_test.zip -- issue is in diurnal.cpp related to `m_colours` having 13 defined colors. Need to handle displaying weekday/weekend schedules with up to 36 values.

[...]

PR #1356, opened 2023-02-18
Addresses SAM issue 1354

sjanzou:
API changes from patch branch


cpaulgilman (commented):
Thanks for tackling this one and for updating the API files for my PR #1352.

1. It looks like the "Max" parameter of the DiurnalPeriod widget for `ur_ec_sched_weekend` changed from 12 to 24, but not for `ur_ec_sched_weekday`. Should these both be 36 for both the weekday and weekend matrix to help address https://github.com/NREL/SAM/issues/28? In my testing the widget works with 36 periods, so I'm not sure it matters what the Max parameter value is.

2. Similarly should invoke.cpp Line 3543 loop 36 times instead of 24?

sjanzou:
> Thanks for tackling this one and for updating the API files for my PR #1352.
> 
> 1. It looks like the "Max" parameter of the DiurnalPeriod widget for `ur_ec_sched_weekend` changed from 12 to 24, but not for `ur_ec_sched_weekday`. Should these both be 36 for both the weekday and weekend matrix to help address #28? In my testing the widget works with 36 periods, so I'm not sure it matters what the Max parameter value is.
> 2. Similarly should invoke.cpp Line 3543 loop 36 times instead of 24?

Yes, definitely! I was going to do a separate pull request for issue 28. The max parameter should set how many colors are added and should limit the max period number.

PR #1358, opened 2023-02-20
2022.11.21.r1.ssc.279 release candidate

sjanzou:


brtietz (commented):
The changes to data_structures.cpp (potentially) casue PySAM to fail during the build. The error is below from running `python setup.py install` on main in the PySAM repository, after running api_autogen in SAM:

image

Are you able to reproduce this issue or are there old files I need to clear out?

sjanzou:
> The changes to data_structures.cpp (potentially) casue PySAM to fail during the build. The error is below from running `python setup.py install` on main in the PySAM repository, after running api_autogen in SAM:
> 
> image
> 
> Are you able to reproduce this issue or are there old files I need to clear out?

@brtietz ,  please revert the change to data structures.cpp. I changed from <int> to <double> (not sure why we are using <int> aor an array type) and worked fine on Mac but I did not have a chance to test on Windows. Also, please do a clean build to make sure the updates for thenJSON forms are included.

brtietz:
@sjanzou That worked for the previous error and I've pushed the reversion. Unfortunately there is a new error:

image

Any ideas on this one or do we need to pull in the CSP folks? I have not had a chance to test on patch yet.

sjanzou:
> @sjanzou That worked for the previous error and I've pushed the reversion. Unfortunately there is a new error:
> 
> image
> 
> Any ideas on this one or do we need to pull in the CSP folks? I have not had a chance to test on patch yet.

@brtietz - that error has been around for a while - when fixing the dGen NaN issue, I had to edit the generated PhysicalTrough.c file manually per the emails last week - I will fix in a separate pull request per the email from @Matthew-Boyd

PR #1372, opened 2023-03-09
merge patch into develop

sjanzou:


tyneises (approved):
@sjanzou Are you running the repo or local version of deploy/test_script_ow.lk? A while ago I modified it so that the ETES, PTES, MSPT, Physical Trough, IPH models don't write all the notices. Here's the script output I get for the ETES model:
image

brtietz (approved):

sjanzou:
> @sjanzou Are you running the repo or local version of deploy/test_script_ow.lk? A while ago I modified it so that the ETES, PTES, MSPT, Physical Trough, IPH models don't write all the notices. Here's the script output I get for the ETES model: image

@tyneises, I am running from SAM-private/deploy/test_script_ow.lk to make sure nothing between SAM and SAM-private is broken....

PR #1446, opened 2023-09-13
Add new BTM battery dispatch method for self-consumption

janinefreeman:
This pull request and the complementary one in SAM add a new battery dispatch methodology for behind-the-meter PV+Battery and BTM battery hybrid systems that allow the battery to be dispatched to maximize self-consumption, which can also be termed "24/7 carbon-free energy" for larger industrial loads. Behind the scenes, this new dispatch option uses the grid power target option with the entire timeseries set to 0, but it is a new dispatch option to enable this functionality for PySAM users as well. Added a few new tests to go along with the new dispatch method.

- [ x] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes
- [x ] Any dependent changes have been merged and published in downstream modules
- [x ] I have checked my code and corrected any misspellings


deploy/runtime/ui/Battery Dispatch Standalone Options BTM.json
```diff
@@ -267,7 +267,7 @@
             "Label": "Battery dispatch choice",
             "Units": " ",
             "Group": "Battery Dispatch",
-            "IndexLabels": "",
+            "IndexLabels": "Peak shaving| Input grid power targets| Input battery power targets| Manual dispatch| Price signal forecast",
```
brtietz, line 270: Self consumption appears to be missing in the index labels.

brtietz (changes requested):
Overall this looks good, two nitpicks:

(1) IndexLabels appears to be missing self consumption

(2) How should the charge from grid checkbox be handled? My impression is it does nothing (which makes sense for the algorithm) but in that case should it be disabled?

deploy/runtime/ui/Battery Dispatch Standalone Options BTM.json
janinefreeman: This dispatch algorithm doesn't make sense for a standalone system, so I didn't add it to this page. Is there a good place for me to add a comment as to why it's missing here? I'm sure others will have this same confusion...

brtietz: Ah, right! The help documentation is likely the right spot for users. For us coders - does near the list of configurations in ssc make sense https://github.com/NREL/ssc/pull/1049/files#diff-ad752cc0b0fdf4cc68f248c6994ea37977572012346c2829d3efb51d61f0f90bR185 ?

brtietz (approved):
Looks good, thank you!

PR #1448, opened 2023-09-14
New Hybrid Configurations

cpaulgilman:
Defaults and UI for new hybrid configurations for Single Owner and Host Developer financial models:

PVWatts Wind Battery Hybrid

PVWatts Wind FuelCell Battery Hybrid

Photovoltaic Wind Battery Hybrid

Generic PVWatts Wind FuelCell Battery Hybrid

Goes with:

* https://github.com/NREL/ssc/pull/1051
* https://github.com/NREL/wex/pull/164


brtietz:
Minor defaults issue: battery dispatch schedules are not set for some configurations:

image

PVWatts-Wind-Battery looks ok but the others are all 1s.

brtietz:
All of the files in api_autogen being deleted suggests some risk to PySAM functionality but I haven't had the chance to test it. Is there a reason those have been deleted? Has PySAM been tested with this branch at all?

sjanzou:
> All of the files in api_autogen being deleted suggests some risk to PySAM functionality but I haven't had the chance to test it. Is there a reason those have been deleted? Has PySAM been tested with this branch at all?

PySAM is way after the Fall Release. The api_autogen will need to be rewritten entirely as discussed previously at the hybrids meetings. The defaults and forms should be made to align with the SAM JSON instead of another variation. PySAM will not run with this branch and has not been tested.

brtietz (commented):
Thanks for fixing the utility rate bug!

Time permitting, could the stuff under src get a pass for commented out code as well? I expect more of the TODOs here are still real, but it would be good to know which sections of commented out code are "reinstate this when we fix ABC" vs "pre-hybrids and no foreseeable use"

Issue #1427, opened 2023-08-08
Code generation for c#


sjanzou:
The generated code calls to the c sharp function bool Log has float instead of the required double declaration for the time argument  which causes an error
<img width="381" alt="image" src="https://github.com/NREL/SAM/assets/6498311/e7363790-b7a2-4e6a-b947-3e168c00553d">

If changed to double, there is no error and the code runs and matches SAM

The generated code should have extension .cs instead of .cpp (both work with csc.exe command line compiler)

Details in attached document
[UserSupportForum2023.8.7_issue_4329.docx](https://github.com/NREL/SAM/files/12289471/UserSupportForum2023.8.7_issue_4329.docx)

cpaulgilman:
https://sam.nrel.gov/forum/forum-sdk/4329

PR #1451, opened 2023-09-19
Fixes issue #1427

sjanzou:
To test:
1.	Create PVWatts/No Finance case in SAM and generate c# code
2.	VS2022 Community version with .NET desktop development added for c# support 
3.	Start VS2022
4.	Tools->Visual Studio Command Prompt
5.	Navigate to where c sharp code generated from SAM within the command prompt
6.	In the command prompt, type “csc untitled.cs”
7.    Check that results match with SAM


cpaulgilman (approved):
Runs and generates results consistent with SAM. Tested with VS Code.

Issue #1452, opened 2023-09-19
Mac version crashing 


sjanzou:
Latest develop branch builds of SAM crashing - release reports 
![image](https://github.com/NREL/SAM/assets/6498311/9e63b405-05bf-4c10-934a-63230ce26f90)

Same hard crash for hybrid and non-hybrid configurations.

PR #1454, opened 2023-09-20
Fix infinite loop that caused issue #1452

sjanzou:
To test:
1. build develop branches on MacOS
2. Run SAM 
3. Create PVWatts / No Financial case
4. Simulate - hard crash

5. Repeat steps 1 through 4 with SAM_1452 branch of SAM and simulation runs without issue


cpaulgilman (approved):
Looks good -- builds and runs on Mac

PR #1462, opened 2023-09-25
Defaults manager updates for hybrid - issue 1461

sjanzou:
Note that for hybrid configurations, the variable to be modified is prepended with the bin_name, e.g. to update degradation for the Wind subsystem in a hybrid configuration, please use "wind_degradation".

If in doubt, use the "Lookup" button on the defaults manager page, e.g.
image

Example to add "wind_degradation" (after I deleted in all configurations using the defaults manager):
image


mjprilliman (approved):
Worked well for me with SAM_1184 changes. Working to fix prechecks to get tests passing.

PR #1463, opened 2023-09-25
Set adj losses to zero for pv+battery cases

brtietz:
Merge into develop losses fix I used during the battery webinar. Emails/SAM meeting feedback confirmed these weren't intentional.


brtietz:
Changed to draft because I remembered that I need to re-run the test results for this type of change. Going to let https://github.com/NREL/SAM/pull/1462 go first, then I'll re-run the tests.

cpaulgilman (approved):
Looks good for PV-battery configs.

All CSP and IPH configs have a adjust_constant = 4 by default except for:

MSPT, None
MSPT IPH, None
MSPT IPH, LCOH

Maybe set those to 4 to be consistent and then confirm with @tyneises and @taylorbrown75?

tyneises:
> Looks good for PV-battery configs.
> 
> All CSP and IPH configs have a adjust_constant = 4 by default except for:
> 
> MSPT, None MSPT IPH, None MSPT IPH, LCOH
> 
> Maybe set those to 4 to be consistent and then confirm with @tyneises and @taylorbrown75?

We ought to consider whether all IPH models should be something other than 4 considering they don't have a power block, but for now I think making everything 4 is a good plan

PR #1466, opened 2023-09-26
Update "Change model..." fixes #1457

### LINKED ISSUE #1457: Change model on develop leaves menus in place
### Describe the bug

The "change model" interface on develop doesn't re-load the menus for the new model's interface. For example, going from a PV+Battery case to a detailed PV only case leaves the battery menus in place.

### To Reproduce

1. Create a PV+Battery model in the beta/on develop
2. Change the model to PV only
3. The battery options are still in the menu

### Expected behavior

Drop downs not relevant to the current model should be removed.

### Screenshots

image

 - OS: Windows
 - Version 2023.9.19/develop

sjanzou:


cpaulgilman (commented):
This works for switching from Detailed PV-Battery / Single Owner to Detailed PV / Battery, but SAM crashes when switching from PVWatts Wind Battery Hybrid / Single Owner to Wind / Single Owner.

brtietz:
> This works for switching from Detailed PV-Battery / Single Owner to Detailed PV / Battery, but SAM crashes when switching from PVWatts Wind Battery Hybrid / Single Owner to Wind / Single Owner.

I get the same results. Interestingly PVWatts Wind Battery Hybrid / Single Owner to Detailed PV / Single Owner works, so it seems like whatever is causing the crash might be more wind related than hybrids related.

cpaulgilman (approved):
This fixes the change model from hybrid to wind issue, and the original #1457.

PR #1475, opened 2023-09-29
Hybrid Results Summary and Cash Flow tabs

cpaulgilman:
Organize metrics tables and autographs on Summary tab for all hybrid configurations

Add cash flow operating cost section headers for hybrid configurations.

Fix issues with results page causing crashes when switching between Results page tabs and state of case changes.


sjanzou (approved):
Reviewed while testing and fixing some exceptions encountered!

Thanks for updating - results look much better!

Issue #1485, opened 2023-10-04
Results button resetting when linked hybrid variables are reloaded


mjprilliman:
### Describe the bug

-After clicking Simulate, if an input page with a dependent input variable (system_capacity, total_installed_cost, etc.) is loaded, the Results button to view previously simulated results is deactivated and cannot be clicked. This is fine functionally but is not immediately clear to the user when it happens. 
### To Reproduce

1. Go to Hybrid case
2. Hit Simulate
3. Click on PV System Design
4. Try to Click Results button next to Simulate

### Expected behavior

-Not clear what the expected behavior should be. Perhaps a Tooltip or a change in color for the button when it is not active.

cpaulgilman:
A possible solution would be to clear results instead of disabling the Results button so that users get some feedback from the message in the Metrics table when they click the Results button:

![image](https://github.com/NREL/SAM/assets/27711933/aa64514a-5927-4e21-ba21-50e5e60ceb52)

This is still a bit confusing, because it isn't clear why the results are cleared, but is better than no feedback.

PR #1498, opened 2023-10-13
addresses SAM issue 1455

sjanzou:
Goes with https://github.com/NREL/wex/pull/165

Sort order for technology tree specified in startup.lk as (note that not all technologies have to be specified):
image

Which results in configuration window display as:
image

If no "TechnologyTreeOrder" configopt specified in startup.lk, the result is how items appear in startup.lk:
image

The financial tree and page order sorting has not yet been implemented. 

Will proceed based on this pull request feedback

### DIFF
--- a/deploy/runtime/startup.lk
+++ b/deploy/runtime/startup.lk
@@ -192,7 +192,7 @@ function setup_pages_battery_dispatch_btm()
 	          [ {'name'='Battery Dispatch Grid Power Targets', 'caption'='Grid Power Targets'} ],
 		      [ {'name'='Battery Dispatch Battery Power Targets', 'caption'='Battery Power Targets'}],
 		      [ {'name'='Battery Dispatch Manual', 'caption'='Manual'}],
-		      [ {'name'='Battery Dispatch Price Signal', 'caption'='Price Signal'} ],
+		      [ {'name'='Battery Dispatch Retail Rate', 'caption'='Retail Rate'} ],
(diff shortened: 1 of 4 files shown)


mjprilliman:
Is there a way to clear the input page selection from the tree when the Results button next to Simulate is pressed? If you do steps:

1. Click Simulate
2. From Results page click an input page to return to inputs.
3. Click Results button to right of Simulate to return to Results
4. Click on still highlighted input page in tree. 

Nothing happens. The easy workaround is to select a different input page. There is also the question of the Results button being disabled, which has it's own issue #1485

cpaulgilman (approved):
Configuration Tree sorting works as advertised in Linux (Mint 21.1).

With sorting in startup:

tech-tree-sorted

Without sorting, Hybrid is at bottom:

tech-tree-not-sorted

Note that in Linux, selected technology is white (invisible) when financing is selected (not a high priority to fix):

tech-tree-invisible-tech

sjanzou:
> Is there a way to clear the input page selection from the tree when the Results button next to Simulate is pressed? If you do steps:
> 
> 1. Click Simulate
> 2. From Results page click an input page to return to inputs.
> 3. Click Results button to right of Simulate to return to Results
> 4. Click on still highlighted input page in tree.
> 
> Nothing happens. The easy workaround is to select a different input page. There is also the question of the Results button being disabled, which has it's own issue #1485

Updated with latest commit - please retest.

sjanzou:
tested RHEL 8.6, MacOS 14.0 and Windows 11

image

PR #1499, opened 2023-10-13
New LK NSRDB download example

cpaulgilman:
This is a new LK sample file that shows how to download a weather file from the NSRDB PSM V3.2.2 TMY API endpoint in response to this Forum post: https://sam.nrel.gov/forum/forum-general/4385.


mjprilliman (approved):
Worked on sam-private build

PR #1513, opened 2023-10-22
Fix SAM #1485

sjanzou:


cpaulgilman (commented):
Functionality is good, but Results button requires two clicks to switch to Results page instead of one.

sjanzou:
> Functionality is good, but Results button requires two clicks to switch to Results page instead of one.

I have tested on Windows 11 and RHEL 8.6 and the results button is only requiring one click for me
image

@mjprilliman , @brtietz  - how many clicks is the results button requiring for you to go to the results page from an input page?

brtietz (approved):
This works for me with one click

cpaulgilman (approved):
Works on Mac with single click.

Issue #1472, opened 2023-09-27
Cannot assign hybrid subsystem degradation in UI


cpaulgilman:
### Describe the bug

AC degradation input is automatically set to value of AC degradation for first subsystem in navigation menu. For example for Generic PVWatts Wind FuelCell Battery configuration, all AC degradation inputs are set to the value of Generic AC degradation.

### To Reproduce

1. Crate a Generic PVWatts Wind FuelCell Battery / Single Owner case.
2. Go to Generic AC degradation page, change to 0.3.
3. Go to PV AC Degradation page, try to change AC degradation.

### Expected behavior

Each subsystem should have separate AC degradation value.

PR #1516, opened 2023-10-23
Fix issue #1472

sjanzou:


mjprilliman (approved):
Can confirm different AC degradation were applied in Generic, PVWatts, and Wind as calculated against annual energy outputs for each.

cpaulgilman (approved):

Issue #313, opened 2020-05-18
Feature request: REopt size for resilience


dguittet:
carrying over request in #310 from @brianschmidly

Let SAM pick the battery size (and solar PV size) based on probability of surviving an outage and then filter the results (in a parametric study) By NPV or savings or simple payback. 

Awaiting response from REopt team on availability of resilience as objective from REopt Lite API.

brtietz:
Some additional issues that I'm going to clean up as a part of the pull request for this issue:

1. The "Get size and dispatch" button _always_ applies the custom dispatch value even if the user selects "no" on the sizing, ask the user about this too.
2. REopt (now?) has inputs to allow for only charging from PV, so we can avoid forcing charge to grid in the API call.
3. The solution to https://github.com/NREL/ssc/issues/923 used "gen" as the variable for the production factor series, but this includes the battery dispatch throwing off the times that the battery should be charging. Swap this over to PV production without battery.
4. https://github.com/NREL/ssc/blob/6c9fda04a7253385ed90b55dab5c3b2788b718d3/ssc/cmod_battery_eqns.cpp#L159 assigns inverter and rectifier efficiencies to the internal efficiency variable, we can reassign these and leave internal efficiency at the default.

PR #1527, opened 2023-10-28
Sam 313 reopt v3 for resilience

brtietz:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes https://github.com/NREL/SAM/issues/313

Pairs with SSC PR https://github.com/NREL/ssc/pull/1078, recommend testing with SAM-private PR https://github.com/NREL/SAM-private/pull/96 as well

To test:

- REopt API call should no longer require grid charging
- New API call should be present on grid outage page for PV-battery configurations
- Sample outage to test with: 
[oct_8th_outage_1wk.csv](https://github.com/NREL/SAM/files/13196765/oct_8th_outage_1wk.csv)

Other notes:
- kwh/kw provided to REopt should always match the SAM case now, so removed the text about lat/lon - this will require a help update
- Reflects V3 updates from SSC pull request

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [x] This change modifies variables in existing compute modules. Please see [Checking for PySAM Incompatible API Changes](https://github.com/NREL/SAM/wiki/Checking-for-PySAM-Incompatible-API-Changes).

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

### DIFF
--- a/deploy/runtime/ui/Battery REopt Optimization.json
+++ b/deploy/runtime/ui/Battery REopt Optimization.json
@@ -399,8 +399,8 @@
         "",
         "};",
         "",
-        "function enable_consider_grid_outage() {\r",
-        "\tenable('reopt_consider_outage',  outage_setup);\r",
+        "function enable_consider_grid_outage() {",
+        "\tenable('reopt_consider_outage',  is_crit() && is_grid_outage());\r",
         "}"
     ]
 }


cpaulgilman (commented):
1. I am getting an intermittent UI callback error -- is this being triggered by an unexpected response from the API? Clicking the button again once or twice after the error seems to get it to work.

```
Could not evaluate callback function:call_reopt->on_change
[103]: error: _s
[103]: error: _s
[5]: eval error in statement list
```

2. Is there a difference in functionality between the REopt button on the Battery Cell and System page and the one on the Grid Outage page? I find the two buttons confusing. If the two buttons do the same thing, I think I prefer a single button on the Battery Cell and System page. A alternate approach would be to have a separate optimization page because it technically affects battery size, dispatch, and grid outages.

brtietz:
1. I saw this while testing as well and think it's an intermittent API error. Do you have any tips on how to propagate the errors at least out to these callback errors? The current error handling looks ok but is clearly missing this one.

2. The battery page does economic-only battery optimization with the PV size fixed, whereas the one on the grid outage page includes the outages and recommends a new size for PV. This is controlled with a boolean passed to invoke.cpp, so we could shift to a dialog choice easily enough if that's less confusing.

cpaulgilman:
> 2\. The battery page does economic-only battery optimization with the PV size fixed, whereas the one on the grid outage page includes the outages and recommends a new size for PV. This is controlled with a boolean passed to invoke.cpp, so we could shift to a dialog choice easily enough if that's less confusing.

I think I prefer a single button on the Battery Cell and System page and a dialog choice if outage is enabled. That will more clearly communicate to experienced users that there is a new feature. We can also add a sentence to the UI text explaining that considering grid outage is an option.

cpaulgilman:
> 1. I saw this while testing as well and think it's an intermittent API error. Do you have any tips on how to propagate the errors at least out to these callback errors? The current error handling looks ok but is clearly missing this one.

Hard to say without knowing what the API is returning. It might be returning a non-JSON file (maybe HTML?) in which case returning a fixed number of characters of the file to display in a UI dialog may be a way to handle it. Or, retrying the API call once or twice if it doesn't return results with an error message.

brtietz:
> > 2. The battery page does economic-only battery optimization with the PV size fixed, whereas the one on the grid outage page includes the outages and recommends a new size for PV. This is controlled with a boolean passed to invoke.cpp, so we could shift to a dialog choice easily enough if that's less confusing.
> 
> I think I prefer a single button on the Battery Cell and System page and a dialog choice if outage is enabled. That will more clearly communicate to experienced users that there is a new feature. We can also add a sentence to the UI text explaining that considering grid outage is an option.

Addressed in latest push.

brtietz:
> > 1. I saw this while testing as well and think it's an intermittent API error. Do you have any tips on how to propagate the errors at least out to these callback errors? The current error handling looks ok but is clearly missing this one.
> 
> Hard to say without knowing what the API is returning. It might be returning a non-JSON file (maybe HTML?) in which case returning a fixed number of characters of the file to display in a UI dialog may be a way to handle it. Or, retrying the API call once or twice if it doesn't return results with an error message.

I _think_ this was due to REopt V3 returning an empty dictionary if no storage is optimal, rather than size_kw = 0. I've pushed a fix to address this, can you let me know if it clears up the issue on your end?

cpaulgilman:
> > > 2. The battery page does economic-only battery optimization with the PV size fixed, whereas the one on the grid outage page includes the outages and recommends a new size for PV. This is controlled with a boolean passed to invoke.cpp, so we could shift to a dialog choice easily enough if that's less confusing.
> > 
> > 
> > I think I prefer a single button on the Battery Cell and System page and a dialog choice if outage is enabled. That will more clearly communicate to experienced users that there is a new feature. We can also add a sentence to the UI text explaining that considering grid outage is an option.
> 
> Addressed in latest push.

This functionality is better. But, after testing, I think a checkbox like "Consider outage in REopt sizing" on the Battery Cell and System page would be better. It would be enabled when grid outage data is specified and disabled (and set to false) otherwise. I can add that if you agree.

cpaulgilman:
@brtietz "Consider grid outage in REopt optimization" checkbox is ready for your review.

brtietz:
> @brtietz "Consider grid outage in REopt optimization" checkbox is ready for your review.

Thank you, the checkbox looks good! I made two modifications:

1. Remove the debug "howdy doody" box
2. Add logic so that if the grid outage gets turned off, the box gets unchecked.

Can you confirm (2) is the desired behavior?

cpaulgilman:
> > @brtietz "Consider grid outage in REopt optimization" checkbox is ready for your review.
> 
> Thank you, the checkbox looks good! I made two modifications:
> 
>     1. Remove the debug "howdy doody" box
> 
>     2. Add logic so that if the grid outage gets turned off, the box gets unchecked.
> 
> 
> Can you confirm (2) is the desired behavior?

Oops, thanks for (1). 

I went back and forth on (2), and ended up deciding not to automatically uncheck the box. The REopt call does not consider outages when the checkbox is either disabled or unchecked, which I think is clear enough, although we are not consistent with this throughout the UI. (In general, it's best not to automatically change inputs without notifying the user.)

I could see a situation where a user is trying different scenarios with and without outages on the Grid Outage page and may not notice that the box gets unchecked on the Battery Cell and System page. Or, they may be manipulating critical load and grid outage inputs in an LK script and not notice the check box being unchecked.

On the other hand, running REopt requires clicking the button on the Battery Cell and System page, so the user *should* see the unchecked box, so I could be convinced that automatically unchecking is the clearest option. You can decide.

cpaulgilman (approved):
I'll approve this pending your final decision on automatic unchecking so you can get this merged when you're ready.

PR #1535, opened 2023-10-31
Fix configuration tree order

cpaulgilman:
Set configuration tree order for performance models to:

Photovoltaic
Energy Storage
Hybrid
Concentrating Solar Power
Industrial Process Heat
Marine Energy
Wind
Fuel Cell - PV - Battery
Geothermal
Solar Water Heating
Biomass Combustion
Generic System


sjanzou (approved):
Thanks for reordering.

Issue #1456, opened 2023-09-20
Version upgrade issues


cpaulgilman:
## Issue 1: UI callback error before versions.lk runs:

1. From Develop version of SAM, open mslf-so-2022-11-21r3.sam from [version-upgrade-test.zip](https://github.com/NREL/SAM/files/12678106/version-upgrade-test.zip). This is a default MSLF / Single Owner case created in SAM 2022.11.21 r3.

2. UI callback error before version upgrade message:

![image](https://github.com/NREL/SAM/assets/27711933/8f9c39d8-08cb-4250-8753-b99448485beb)

## Issue 2: SAM crashes after opening case with configuration that no longer exists.

1. From Develop version of SAM open mslf-comm-2022-11-21-r3.sam (in version-upgrade-test.zip linked above). This is a default MSLF / Commercial case created in SAM 2022.11.21 r3. The Commercial financial model is not an option for MSLF in the Develop build.

2. SAM crashes on startup.

cpaulgilman:
Note that @tyneises and @taylorbrown75 and I discussed removing Merchant Plant from all CSP configurations, which would require fixing Issue 2 above.

brtietz:
The simulation fails for PV  systems with "estimate subarray 1 configuration" checked:

![image](https://github.com/NREL/SAM/assets/5530592/80c6692c-ffde-455e-9584-21d4c61e5edd)

See attached SAM file, created with 2022.11.21: 
[version_upgrade_sizing_tests.zip](https://github.com/NREL/SAM/files/12751390/version_upgrade_sizing_tests.zip)

Estimate subarray 1 configuration works fine in each version independently, so suspecting this is a version upgrade issue.

sjanzou:
> The simulation fails for PV systems with "estimate subarray 1 configuration" checked:
> 
> ![image](https://user-images.githubusercontent.com/5530592/271337192-80c6692c-ffde-455e-9584-21d4c61e5edd.png)
> 
> See attached SAM file, created with 2022.11.21: [version_upgrade_sizing_tests.zip](https://github.com/NREL/SAM/files/12751390/version_upgrade_sizing_tests.zip)
> 
> Estimate subarray 1 configuration works fine in each version independently, so suspecting this is a version upgrade issue. Additionally, unchecking the box then attempting the version upgrade fails, so there might be a larger problem with detailed PV subarrays here.

[...]

PR #1554, opened 2023-11-09
Addresses issues 1 and 2 in #1456

sjanzou:
Please test with files in #1456


cpaulgilman (approved):
Looks good: Old files with removed configurations result in error message instead of crash, and version upgrade no longer triggers UI callback errors. Also tested module strings issue.

PR #1556, opened 2023-11-09
Remove Windows "quick build" steps from SAM readme

cpaulgilman:
Better to maintain a set of complete instructions for each operating system than to duplicate and maintain these quick steps for Windows only.


brtietz (approved):
Thanks for removing these. Given that cmake and google test were missing, I would be surprised if these could be used successfully.

PR #1586, opened 2023-11-17
Remove wxWidgets version number from SAM readme.

cpaulgilman:


brtietz (approved):
Thanks, agreed that it's reasonable to have people go to the build instructions to find the current versions.

Issue #1583, opened 2023-11-16
PVWatts UI equations not updating


brtietz:
### Describe the bug
The PVWatts financial UI equations (to calculate total installed cost, construction finance cost, etc) are not automatically running when a case is opened or when the pages are loaded. The same pages for detailed PV seem to work.

### To Reproduce

1. Check out branch 2023_defaults_updates to update component costs
2. Open a PVWatts case
3. Calculated values do not update

![MicrosoftTeams-image (5)](https://github.com/NREL/SAM/assets/5530592/35ecd50a-4435-4467-acab-b8bb1bb6fbe8)

Note how the component costs times the DC capacity does not equal the calculated value on the right.

### Expected behavior

Behavior should match detailed PV with automatic calculated updates.

PR #1587, opened 2023-11-17
Fixes issue #1583

sjanzou:
Fixes issue #1583 and all files from #1456 work with this pull request

image


brtietz (approved):
This works, thank you!

Issue #1596, opened 2023-11-19
Update lk functions to recalculate equations for test scripts


sjanzou:
1. Modify [overwrite_all_defaults.lk](http://overwrite_all_defaults.lk/) script to accommodate hybrids changes
2. Ac_nameplate for hybrids vs pvwatts (discussed with Brian): default is 0, Brian thinks equations should overwrite the default to shift+F10 works in the future

The issue involves explicitly recalculating equations in SAM configurations since this functionality was removed from Case::SetConfiguration per #1456

sjanzou:
Closes when 2023_defaults_updates is merged into develop per #1597

PR #1597, opened 2023-11-19
fixes issue #1596 - update calculated value using overwrite_all_defaults.lk

sjanzou:
Should also fix https://github.com/NREL/ssc/issues/1092


brtietz (approved):
Fantastic! Great to have the script working again.

The changes to startup.lk for https://github.com/NREL/ssc/issues/1092 work as well. I had to make some additional changes to the defaults to make everything load properly (both hybrid_load and battery_load are required) and have pushed those here.

PR #1602, opened 2023-11-20
Version upgrade updates

janinefreeman:
Updates the version upgrade script to make the language clearer for the adjust variable changes (from table with semicolons to separate variables with underscores) and fixes the logic for upgrading the "enable custom time period" losses in that widget, which were not upgrading correctly in the current dev version.

Please specifically test enabling a custom time period loss in SAM 2022.11.21 and upgrading that file, and check that the loss upgrades correctly.

### DIFF
--- a/deploy/runtime/versions.lk
+++ b/deploy/runtime/versions.lk
@@ -68,7 +68,11 @@ version_upgrade{'2023.9.19'} = define() {
 		message('Availability losses variable format changed from single table ("adjust") to separate variables: adjust_constant (number), adjust_en_timeindex (number), adjust_timeindex (array), adjust_en_periods (number), adjust_periods (array).');
 		value('adjust_constant', x.constant);
 		value('adjust_en_periods', x.en_periods);
-		value('adjust_periods', x.periods); 
+		//assigned periods are a matrix [[x,y,z]], unassigned periods are [1]. the latter must be reformatted to [0,0,0] to simulate correctly.
+		p = x.periods;
+		p1 = p[0];
+		if (typeof(p1) == 'number') {p = [[0,0,0]];}
+		value('adjust_periods', p); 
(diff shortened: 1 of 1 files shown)


deploy/runtime/versions.lk
```diff
@@ -81,86 +87,101 @@ version_upgrade{'2023.9.19'} = define() {
...
+		value('dc_adjust_en_timeindex', x.en_timeindex);
+		value('dc_adjust_timeindex', x.timeindex);
+		
+		//AC adjust factor
+		x = oldvalue('adjust');
+		message('AC losses variable format changed from single table ("adjust") to separate variables: adjust_constant (number), adjust_en_timeindex (number), adjust_timeindex (array), adjust_en_periods (number), adjust_periods (array).');
+		value('adjust_constant', x.constant);
+		value('adjust_en_periods', x.en_periods);
+		p = x.periods; //this logic works for both blank period arrays and for completed ones
+		value('adjust_periods', p);
```
mjprilliman, line 110: adjust_periods_dc fails on simulate when upgrading the default detailed PV case (without changing anything)

image

mjprilliman:
Will close #1604

deploy/runtime/versions.lk
janinefreeman: I can't replicate this on my machine. However, what is getting loaded in seems to have changed and is no longer working for me. Possibly a wxwidgets issue? Let's talk more at our working session...

mjprilliman:
[version-upgrade-testing-csp.zip](https://github.com/NREL/SAM/files/13506311/version-upgrade-testing-csp.zip)
My goal is to run the cases in these two files through the upgrade script and have them simulate without any error.

mjprilliman (approved):
The last commit fixed the issues I was finding on the first review. Thanks for working through this!

PR #1581, opened 2023-11-16
2023 defaults updates

brtietz:
Update defaults for the 2023 release. Changes include:

1. Update PV and battery+x costs to align with 2023 PV+battery benchmark
2. Update standalone battery and wind costs to align with 2023 ATB
3. Update financials to align with 2023 ATB
4. Fix https://github.com/NREL/SAM/issues/1471
5. Update PPA prices to produce positive NPVs with the above (at least for single owner)
6. Update computed values for PySAM with automated override script

All references pushed to SAM-documentation.

The following have not been updated:
1. Host developer sizes (pending fix to https://github.com/NREL/ssc/issues/1092)
2. PySAM defaults/API autogen files

PR #1605, opened 2023-11-28
Fixes #1594 for hybrids and non-hybrids

### LINKED ISSUE #1594: Parametrics import function doesn't work
The import function in parametrics is broken.

Steps to reproduce:
1. Open a case (I've tested with Detailed PV + Battery/Commercial and PVWatts/None)
2. Create a parametrics table (as simple as can be)
3. Export the table as a CSV
4. Attempt to import the CSV table
5. SAM hard crashes

The import function is broken in the current release as well, but gives an error message "Error: could not identify parametric variable run" rather than crashing. Given that we haven't gotten complaints about this so far, I'm going to give it a lower priority and assign it to the patch.

sjanzou:


sjanzou:
Note that test script will fail until #1581 is merged into develop and develop merged into SAM_1594 branch of SAM

PR #1606, opened 2023-11-28
Update CEC Modules for Fall 2023 release

cpaulgilman:
Update CEC module library

* New CEC Modules.csv file
* Files used for generating library added to samples/CEC Module and Inverter Libraries

**Note** Merge this before merging inverter update.


mjprilliman (approved):

cwhanse:
Q: the 1344 lines in the "bad module" file, these entries in the CEC list where the model parameters could not be determined, is that correct?

cpaulgilman:
> Q: the 1344 lines in the "bad module" file, these entries in the CEC list where the model parameters could not be determined, is that correct?

Yes. That file includes the inputs and the error messages that we can use to troubleshoot the parameter solver.

PR #1610, opened 2023-11-29
Fixes #1577

### LINKED ISSUE #1577: Intermittent crashes with Inputs Browser
### Describe the bug

-SAM seems to consistently crash with Inputs Browser opening or already opened + multiple hybrid cases
-

sjanzou:
Some comparison limitations
image

Using test file with hybrid and non-hybrid cases and note that non-hybrid cases have variables prepended with technology name (e.g. pvwatts_tilt)
[SAM_1577.zip](https://github.com/NREL/SAM/files/13499793/SAM_1577.zip)

### DIFF
--- a/src/variablegrid.cpp
+++ b/src/variablegrid.cpp
@@ -151,7 +151,8 @@ void VariableGridData::Init()
 			ConfigInfo* ci = m_cases[iCase]->GetConfiguration();
 			for (size_t iVarTable = 0; iVarTable < m_var_info_lookup_vec[iCase].size(); iVarTable++) {
 				// iterate over case vartables
-				wxString prepend = ci->Technology[iVarTable].Lower() + "_";
+//				wxString prepend = ci->Technology[iVarTable].Lower() + "_";
+				wxString prepend = ci->Simulations[iVarTable].Lower() + "_";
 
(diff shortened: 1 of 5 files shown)


src/variablegrid.cpp
```diff
@@ -151,8 +151,8 @@ void VariableGridData::Init()
 			ConfigInfo* ci = m_cases[iCase]->GetConfiguration();
 			for (size_t iVarTable = 0; iVarTable < m_var_info_lookup_vec[iCase].size(); iVarTable++) {
 				// iterate over case vartables
-				wxString prepend = "";
-				if (ci->Technology.size() > 1) prepend = ci->Technology[iVarTable].Lower() + "_";
+				wxString prepend = ci->Technology[iVarTable].Lower() + "_";
```
mjprilliman, line 154: Think there's still an issue with some of the prepends having spaces in them, which is causing crashes when opening another case with Inputs Browser open: 

image

image

mjprilliman: Issue seems consistent when you open another case with inputs browser open for these prepends with spaces. Might need to do a find and replace for " " or use different naming scheme.

mjprilliman (commented):
Still having a few issues with spaces in the prepend names.

mjprilliman (approved):

PR #1614, opened 2023-11-30
Fix #1611 issue updating adjust in version upgrader

### LINKED ISSUE #1611: Upgrading "adjust" variable doesn't work
"Custom periods" in the `adjust` variable are not upgrading correctly. The current version upgrade script on the version-upgrade-updates branch has a debugging msgbox showing the issue. There seems to be an issue switching between different matrix types, [ [x,y,z] ] and [ [x], [y], [z] ] for the `adjust_periods` variable.

Using the attached .sam file (created in SAM 2022.11.21), see that the "custom periods" loss is not upgraded correctly.
[upgrade.zip](https://github.com/NREL/SAM/files/13506608/upgrade.zip)

Once this is fixed, code needs to be updated in the version upgrader script in 3 places, currently lines 71, 104, and 114, which is all the places that the `adjust_periods` variable is set.

sjanzou:
Issue was that version-upgrade-updates does not have the fix for issue 1608 merged into develop CaseWindow::UpdateConfiguration
image
– so that after upgrading, the inputs are overwritten with the defaults. 

Update casewin.cpp with changes from develop and test – working and matching (2022.11.21 on left and updated on right - this branch)
image

with upgrade script
image

### DIFF
--- a/deploy/runtime/versions.lk
+++ b/deploy/runtime/versions.lk
@@ -68,13 +68,7 @@ version_upgrade{'2023.9.19'} = define() {
 		message('Availability losses variable format changed from single table ("adjust") to separate variables: adjust_constant (number), adjust_en_timeindex (number), adjust_timeindex (array), adjust_en_periods (number), adjust_periods (array).');
 		value('adjust_constant', x.constant);
 		value('adjust_en_periods', x.en_periods);
-		p = x.periods; 
-		//value('adjust_periods', p); - this no longer works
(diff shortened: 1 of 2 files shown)


mjprilliman (commented):
Seems to fix the issue we were discussing yesterday. I am still having some trouble with the custom period updates but not sure if it is just me. 

Steps: 
1. Create a Detailed PV - LCOE calculator in the 2022 release
2. Save that file, close (don't change anything)
3. Open that case in this branch
4. Click Simulate
5. adjust_periods is shown as [1] in inputs browser, upgrade notes. 
image

PR #1615, opened 2023-11-30
Hybrids defaults updates

brtietz:
# Pull Request Template

## Description

Finish hybrids defaults updates now that the dust has settled on other merges. Update:

- Sizing
- PPA prices
- battery dispatch

Also includes shift-F10 updates from other PRs this week.


deploy/runtime/defaults/PVWatts Wind Battery Hybrid_Host Developer.json
```diff
@@ -55917,7 +55918,7 @@
         "ppa_escalation": 1.0,
         "ppa_multiplier_model": 0.0,
         "ppa_price_input": [
-            0.187
+            0.153
```
mjprilliman, line 55921: General comment/question: solar resource is in Phoenix while wind resource is in Wyoming. Do we want to do something with that or just keep the existing defaults from the individual technologies?

PR #1624, opened 2023-12-04
Update default hybrids wind resource file to Eastern AZ

brtietz:
Update the hybrids wind resource default location to Eastern AZ to be closer to solar resource in Phoenix.


cpaulgilman (approved):
Looks good for all wind hybrid configs

Issue #1618, opened 2023-12-01
Values reset to defaults when using change model


brtietz:
### Describe the bug

When using the "change model" interface on the 2023.12.01 release candidate, values are reset to the defaults.

### To Reproduce

1. Open a model (example: PV + Battery)
2. Change a value (example: 20 year analysis period)
3. Use the "change model" interface to change to a related model (example: generic battery)
4. View the changed values and see the defaults.

### Expected behavior

Values that are a shared by the related models (analysis period, battery dispatch choice, etc) should be retained through the change model process.

### Operating System and Version

 - OS: Windows
 - Version 2023.12.01

PR #1638, opened 2023-12-06
Fixes issue #1618

sjanzou:
Note that "Change model..." does not preserve changes between hybrid and non-hybrid configurations.


cpaulgilman (approved):
Tested changing financial model inputs when switching between non-hybrid performance models.

I added a note to description of Change Model feature in Help:

image

Issue #1630, opened 2023-12-05
Inputs Browsers not working for some technology models


tyneises:
E.g. CSP, Marine:
![image](https://github.com/NREL/SAM/assets/27820293/372d463a-c244-4857-96f6-4f1b94110676)

tyneises:
Inputs Browser still not working for Case Comparison *across technologies*. Except common inputs between Trough and Tower to show up in Different or Equal values. E.g. "ppa_price_input"
Show Different Values:
![image](https://github.com/NREL/SAM/assets/27820293/f0897318-d7c1-436e-ba26-4f2f3ff6fb11)

Show Equal Values:
![image](https://github.com/NREL/SAM/assets/27820293/a05416f2-7cc7-4071-b79c-69b1aa99d1c1)

Show All Values:
![image](https://github.com/NREL/SAM/assets/27820293/c955bf7f-e3be-411e-98c7-20604a822266)

![image](https://github.com/NREL/SAM/assets/27820293/179f273f-8098-4ed6-9560-e8dcea72338d)

sjanzou:
[...]

PR #1646, opened 2023-12-08
Fix issue #1618

sjanzou:
Would like to have additional functionality in inputs browser to better compare across all technologies including hybrids...

Test project file
[SAM_1630.sam.zip](https://github.com/NREL/SAM/files/13613578/SAM_1630.sam.zip)


mjprilliman (approved):
Is it the first cmod in the config that gets printed before the ':'? That's a little confusing for wave as wave_file_reader comes before mhk_wave, but that can be discussed more later. The example file works for me.

brtietz (approved):
Works for me!

tyneises (commented):
There are some remaining problems in #1630

Issue #1314, opened 2023-01-10
No error/notice after entering invalid weather file in parametric tool


williamhobbs:
### Describe the bug

I’ve noticed that if I enter an invalid weather file name in the parametric tool (for example, generating a range of weather file names in Excel and pasting them into SAM, while accidentally including a name that doesn’t exist) or if I enter the name of a weather file that doesn’t work properly, I can still complete a simulation run. SAM appears to default to the weather file selected in the Location and Resource tab without notifying the user. This can cause problems by leading the user to think they are getting valid results for some other weather file, and it would be preferred to produce an error, or at least a warning/notice.  

### To Reproduce

1. Setup a PV model
2. Go to the parametric tool, add location/resource as an input, and any output of your choosing.
3. Click and drag to select the range of weather/location files and copy it
4. Paste the weather/location files into Excel or similar
5. Modify one or more weather file names to mess it up
6. Copy/paste back into SAM and run simulations
7. Note that the simulations complete without errors/notices, and that any case/run with an invalid weather file name completes by using the weather file selected in the Location and Resource tab

### Expected behavior

An error or notice that an invalid weather file has been entered.

### Screenshots
![image](https://user-images.githubusercontent.com/45701090/211675780-18c8a66c-f9c8-4259-8ce3-17b4a072cf2b.png)

### Operating System and Version

 - OS: Windows
 - Version: SAM 2022.11.21

PR #1710, opened 2024-02-26
Addresses SAM issue #1314

sjanzou:
Test per https://github.com/NREL/SAM/issues/1314
Note that equations on Solar Resource Data were updated to fail the simulation if the solar_resource (library value) is blank.


cpaulgilman (approved):
This fixes the issue by displaying an error message about invalid weather file.

Copying and pasting to/from Excel is a little awkward. Seems like tab-delimited values instead of comma-delimited would be better for copy/paste, but that's a separate issue.

PR #1718, opened 2024-02-28
JSON code generator for hybrids - addresses SAM issue 1515

sjanzou:
Goes with ssc pull request 1139

Check out TEST_F(CmodHybridTest, CodeGeneratorPVWattsv8WindBatterySingleOwner) test in ssc that uses PVWatts_Wind_Battery_Hybrid_Single_Owner.json file generated from SAM.


cpaulgilman (approved):
Looks good -- inputs are grouped by technology in JSON, and SSC test passes.

This is different from PySAM JSON, and from JSON generated by Shift-F5 (save case inputs as JSON) and Ctril-F7 (save case inputs as JSON for SSC). Seems like we should consolidate these different save-inputs-as-JSON methods.

PR #1737, opened 2024-03-11
Remove PySAM set_data_ptr

dguittet:
Minor update to sscapi needed for PySAM Hybrids, to remove unused PySAM function


sjanzou (approved):
All projects building and SAM running on Windows!
image

Issue #1362, opened 2023-02-23
Disable or zero-out uncertainty variables


williamhobbs:
### Is your feature request related to a problem? Please describe.

I noticed it isn’t possible to set the uncertainty to zero for any of the variables. For example, I get an error if I use a normal distribution with sigma=0 for row shading loss. 

### Describe the solution you'd like

It would be nice to have a checkbox to enable/disable one or more variables for the uncertainty analysis, or to be able to set them to zero, but I think the checkbox would be preferred. 

A comment from @cpaulgilman in our email exchange: "I think a good way to implement that would be as you suggest... as a check box so you can turn off a distribution without losing its settings and then turn it back on."

### Describe alternatives you've considered

It seems that the only option is to make the uncertainty ranges very very small. 

### Additional context

Here's a mock-up of what I have in mind:

![image](https://user-images.githubusercontent.com/45701090/220975151-43ba2230-8e50-4ffb-b232-1237f51b32ea.png)

PR #1753, opened 2024-04-08
Address SAM issue #1362

sjanzou:
https://github.com/NREL/SAM/issues/1362

To test check out SAM_1362 branch of SAM and patch branches of other repos and build.
1. Create a PV/No financial case
2. Click PV Uncertainty at bottom left 
3. run simulation
4. click checkboxes to enable and disable the uncertainty sources
5. Save and load
6. Load project files from previous versions

image


cpaulgilman:
One small nit to pick: If you clear all the check boxes, you get an LHS error. Might be better to generate a UI error like "Uncertainty simulations require at least one distribution."

image

Issue #1734, opened 2024-03-08
P50 / P90 simulations crash for PV Battery configurations


cpaulgilman:
### Describe the bug

SAM crashes when you run a P50 / P90 simulation for a PV Battery configuration.

P50 / P90 is replaced by Uncertainty for Detailed PV configurations, so maybe needs to be done for PV Battery configurations as well.

### To Reproduce

1. Create a PV Battery / Commercial case.
2. On the PV, Solar Resource page enter any location name, change **Default TMY file** to **Download for all years**, and click **Download and add to library** (skip this step if you already have a folder with multiple weather files).
3. Click **P50 / P90** under the Simulate button.
4. Click the ellipses button **...*** to select the weather file folder with multiple files.
5. Click **Run P50/P90 simulations**.
6. SAM crashes.

### Operating System and Version

 - Windows 11
 - SAM 2023.12.17 r1

cpaulgilman:
P50 / P90 simulations do not seem to be working for any technology:

* PV Battery
* Wind
* Solar Water Heating
* All CSP

Error triggered in `Simulation::Override()` function:

![image](https://github.com/NREL/SAM/assets/27711933/9789fabf-d31f-4faf-ab86-1975dddbe273)

Weather files for testing:

Solar: [phoenix_az.zip](https://github.com/NREL/SAM/files/14844062/phoenix_az.zip)

Wind (old SRW format): [wind-P50P90.zip](https://github.com/NREL/SAM/files/14844072/wind-P50P90.zip)




Solar:

Issue #1681, opened 2024-01-24
P50/P90 simulations do not work for wind


cpaulgilman:
P50/P90 simulations is checking for solar resource files instead of wind resource files.

```
Invalid File P50/P90 simulations do not work with the file extension. Only csv, srw, smw, epw amnd tm2 file extensions are supported. please remove any files with invalid extensions from the weather file folder.
```

See https://sam.nrel.gov/forum/forum-general/4503-p50-p90-simulation-error for description and set of wind resource files for testing.

cpaulgilman:
May be related to https://github.com/NREL/SAM/issues/1734

sjanzou:
Zip file of weather resources not available at https://sam.nrel.gov/forum/forum-general/4503-p50-p90-simulation-error.html#12846

PR #1754, opened 2024-04-08
Address and test both #1681 and #1734

sjanzou:
Test with P50/P90 weather files from https://github.com/NREL/SAM/issues/1734:

[phoenix_az.zip](https://github.com/NREL/SAM/files/14903980/phoenix_az.zip)

[wind-P50P90.zip](https://github.com/NREL/SAM/files/14903996/wind-P50P90.zip)

PV-Battery / Commercial with phoenix_az.zip weather files, you should get P50/P90 results:
image

Wind / No financial using wind-P50P90.zip weather files, you should get P50/P90 results:
image


mjprilliman (approved):
Worked for: 
- PV Battery
- Wind
- Solar Water Heating
- Physical Trough

cpaulgilman (approved):
Also works for solar water heating

Issue #1706, opened 2024-02-22
Defaults manager crashes when querying "fuelcell_dispatch_excl"


cpaulgilman:
### Describe the bug

This issue only affects SAM developers.

### To Reproduce

1. Start SAM and open the SAM Development Environment (Shift-F7).
2. Go to Defaults Manager.
3. Right-click list of configurations and  select **Check all**.
4. Type "fuelcell_dispatch_excl" for variable name and click **Query**.

![image](https://github.com/NREL/SAM/assets/27711933/33ba0eca-10c0-4171-8d12-2536c48a0fe9)

Issue on Line 523 of defmgr.cpp:

![image](https://github.com/NREL/SAM/assets/27711933/ed4b60e9-33d8-4732-a1c4-9dd389c7d638)

### Operating System and Version

 - SAMOS from Patch branch 2/22/2024

PR #1755, opened 2024-04-10
Address SAM issue #1706

sjanzou:
Test per step at #1706 

Note to find in hybrid configurations - precede the search term with the vartable as returned by the query 
image

Note difference for hybrid configurations containing fuel cells and non-hybrid configurations containing fuel cells
image


cpaulgilman (approved):
Thanks for the fix and reminder about querying variable names for hybrid configs.

Issue #1563, opened 2023-11-10
OLE Automation Error in Range


sjanzou:
PV-Battery Commercial configuration
Send to Excel with Equations
![image](https://github.com/NREL/SAM/assets/6498311/7c7129d4-0528-4d20-9f40-bfb670d8daba)

Resulting workbook updates and then returns
![image](https://github.com/NREL/SAM/assets/6498311/36346b27-b65a-47c7-9190-e7e2a36b368b)

and results do not match
![image](https://github.com/NREL/SAM/assets/6498311/25bd7b4a-b0fa-40b2-8f45-3c113af442fd)

The OLE Automation error is from automtn.cpp in src/msw/ole folder in wxWidgets:
![image](https://github.com/NREL/SAM/assets/6498311/2870d8af-4877-4654-a8cc-b63a081434ae)

For additional details see https://github.com/NREL/SAM/pull/1545

PR #1762, opened 2024-04-17
Address issue #1563

sjanzou:
Improved error handling when setting named range values in Excel workbooks with equations
image

To test, run the test file 
[PV_Batt_Com.zip](https://github.com/NREL/SAM/files/15009148/PV_Batt_Com.zip)

in the current release and receive the OLE errors
image

Build SAMOS with the patch branches of lk, wex, and ssc and the SAM_1563 branch of SAM and open and run the the file and "Send to Excel with equations" and  receive no OLE errors and metrics that match between the Excel workbook with equations and SAM
image


cpaulgilman (approved):
Looks good -- I'm not seeing OLE Automation errors for the PV Battery / Commercial config.

PR #1763, opened 2024-04-18
Add example code and data for adding snotel data to a weather file

brtietz:
Add a jupyter notebook that translates from SNOTEL data formats (daily data and inches) to SAM format (hourly data and cm). Include example data and a readme


cpaulgilman (approved):
Thanks for this addition to the samples. Looks good. 

"SNOTEL" is an unfortunate name.

mjprilliman (approved):

PR #1779, opened 2024-05-29
Fix base branch in Github CI

dguittet:
Use base (destination) branch for `pull_request`, but try to find matching branch for `push`


brtietz (approved):
Looks great, thank you! Do any other repos need a similar treatment?

dguittet:
@brtietz Yes, wex has this already for patch. I'll set up the same ci.yml change for develop

Issue #1417, opened 2023-06-13
LHS (PV Uncertainty) failing on MacOS 


sjanzou:
2022.11.21.r3.ssc.280 

Issue reported on SAM forum: https://sam.nrel.gov/forum/forum-general/4253

Using attached project file and weather files.
[2023.6.12_issue_4253.zip](https://github.com/NREL/SAM/files/11731721/2023.6.12_issue_4253.zip)

PR #1782, opened 2024-06-03
Update lhs.bin and stepwise.bin to address #1417

sjanzou:
Review by downloading and testing stochastic simulations using
https://samrepo.nrelcloud.org/beta-releases/sam-beta-macos-2024-6-2.pkg

also, see https://github.com/NREL/SAM-private/pull/108


cpaulgilman (approved):
Looks good! Tested PV uncertainty with SAMOS build from this branch and stochastic with Beta build.

PR #1785, opened 2024-06-10
Rerun defaults on Windows for failing GitHub Actions

sjanzou:


brtietz (approved):
Looks like these are expected changes due to https://github.com/NREL/ssc/pull/1164? If so, looks good.

sjanzou:
> Looks like these are expected changes due to [NREL/ssc#1164](https://github.com/NREL/ssc/pull/1164)? If so, looks good.

Yes, that looks correct - the defaults were not updated so GitHub Actions failed for a while...

Issue #1636, opened 2023-12-05
Scripting doesn't work for Hybrid Cases


dguittet:
### Describe the bug

In a SAM hybrid case, the scripting environment cannot access any of the case's variables.

### To Reproduce

1. Open a SAM hybrid case and a new scripting window
2. Try to access a variable

### Expected behavior

Should see a number or some value, but you'll get <null>

PR #1800, opened 2024-06-19
Address hybrid SAM lk scripting issues in #1636

sjanzou:
To test:
1. check out SAM_1636 branch of SAM and patch branches of other repos

2. open the project file and lk script file in the zip archive 
[SAM_1636.zip](https://github.com/user-attachments/files/15898306/SAM_1636.zip)

3. run the script and select "tilt" from input variable dialog in the hybrid
image
and non-hybrid configurations 
image

Expect results - hybrid configuration
image

Expect results - non-hybrid configuration
image

### DIFF
--- a/src/script.cpp
+++ b/src/script.cpp
@@ -138,7 +138,7 @@ static void fcall_varinfo( lk::invoke_t &cxt )
 
 static void fcall_selectinputs( lk::invoke_t &cxt )
 {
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
Looks good for hybrids variables, and thanks for introducing me to the select_inputs() function. Even an old dog can learn new tricks.

Issue #1634, opened 2023-12-05
ssc_auto_exec not using var table "required_if"


tyneises:
When using ssc_auto_exec to call a cmod, we use ‘SIMULATION_PARAMETER’ to designate variables that the UI doesn’t provide to the cmod. It appears that all the variables that don’t have SIMULATION_PARAMETER are required to be in the UI, even if “required_if” is false. 

For example, the following SSC branch currently contains exactly one new line of code in the mspt cmod that adds a new input var, “is_dispatch123”, that i) is not required, ii) does not have ‘SIMULATION_PARAMETER’ in UI Hints, iii) is not defined in the UI, iv) is not used anywhere in the cmod. The MSPT model will fail on open because it uses ssc_auto_exec on MSPT System Design. https://github.com/NREL/ssc/tree/add-temp-mspt_cmod_var

tyneises:
This SAM issue (probably) requires this fix: https://github.com/NREL/SAM/issues/1464

tyneises:
@sjanzou pull request #1702 enables auto_exec to ignore blank required_if entries (i.e. ""). Is it possible for auto_exec to use conditional expressions for required_if?

I'm trying to remove the Financial Parameters page for the No Financial and LCOH options for the Physical Trough IPH model, but I need to keep the inputs in the cmod for other technology models. When I try to make the inputs from the Financial Parameters page dependent on the value of csp_financial_model, the auto_exec call back fails:
https://github.com/NREL/ssc/tree/remove-financing-some-iph
https://github.com/NREL/SAM/tree/remove-financing-some-iph

tyneises:
I'm replacing an old comment that I just deleted to try to clarify the problem and give simple examples in the this comment... 

@sjanzou pull request #1702 **makes** auto_exec ignore blank required_if entries (i.e. ""). I think ideally it would check for those variables but not require them. 

More importantly, auto_exec still cannot use conditional or default required_if. For example, in the cmod below, I change the required_if field for SSC_INPUT "is_dispatch123" in two ways:

1. "is_dispatch = 1"
2. "?=0"

In both cases, the auto_exec callback from the System Design page fails because it can't find "is_dispatch123" in the UI even though it isn't required.

My wish is that the auto_exec would have the same cmod parsing functionality as clicking Simulate in the UI.

https://github.com/NREL/ssc/tree/add-temp-mspt_cmod_var

PR #1801, opened 2024-06-20
Address SAM issue #1634 with ssc_auto_exec

sjanzou:
The fundamental issue for the ssc_auto_exec in the EqnDatabase is that the EqnDatabase is constructed at load up on a page by page basis and has no idea on which UI variables will be available in whatever configuration the page is used. The ssc_auto_exec was constructed to avoid having to manually set each variable when the compute module referenced was called. Initially, any variables that were not to be set by the ssc_auto_exec function were designated by the UIHINT=”SIMULATION_PARAMETER” which is still in effect.

For SAM issue 1634, there is no way for ssc_auto_exec to have the same functionality as clicking the Simulate button in the UI. This is because ssc_auto_exec in constructed at page loadup when SAM starts and has no concept of a case or a configuration. That is, the list of UI variables is not available when the ssc_auto_exec function is evaluated.

To address SAM issue 1634 as much as possible, no UI variable will be used for any ssc variable that has the “REQUIRED_IF=*”. If the ssc variable is not required or has a default or a different REQUIRED_IF value, then ssc_auto_exec will skip the assignment to a UI variable. 

Tested with add-temp-mspt_cmod_var branch of ssc - no equation errors
image


sjanzou:
@tyneises, can you take a look at the failing MSLF configurations?

sjanzou:
MSLF configurations fail because REQUIRED_IF is not set for all required variables.
add-temp-mspt_cmod_var and patch branch work fine for MSPT configurations.
image

sjanzou:
Added fcall_var_exists to facilitate desired functionality of #1634 

Working and running for add-temp-mspt_cmod_var and patch branches of ssc as requested.

@tyneises, @taylorbrown75 please review and let me know if you have any questions.

taylorbrown75 (approved):
Logic is clear and works well. I produced the following table showing performance depending on required_if type and whether the variable is defined in the UI. Works as described. 

image

Seems we can stop using SIMULATION_PARAMETER? Conditional required_if works with sim_type. (i.e. use REQUIRED_IF="sim_type=1")

sjanzou:
> Logic is clear and works well. I produced the following table showing performance depending on required_if type and whether the variable is defined in the UI. Works as described.
> 
> image
> 
> Seems we can stop using SIMULATION_PARAMETER? Conditional required_if works with sim_type. (i.e. use REQUIRED_IF="sim_type=1")

@taylorbrown75 thanks for the detailed review!
@tyneises can we get stop using SIMULATION_PARAMETER as @taylorbrown75 suggests?

taylorbrown75:
Updated table to include conditional required_if case.

image

PR #1805, opened 2024-07-09
1477 itc incentive interactions

brtietz:
Companion PR to https://github.com/NREL/ssc/pull/1178

As of 7/9/24, just updates test results. Need to update cash flow and send to Excel


brtietz:
Closing since we decided to keep the reserves for patch.

PR #1807, opened 2024-07-16
Tidal turbine power curve generation

mjprilliman:
## Description

-Modify ssc equation generating tidal turbine power curves to have rated power based on capacity factor calculations rather than an input power
-Show calculated power per rotor
-For patch 2, using a fixed capacity factor of 0.3 in ssc; for release expose capacity factor as an additional UI input
-todo for release: rename UI variables for improved clarity. 

-See email for test materials

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This change modifies variables in existing compute modules. Please see [Checking for PySAM Incompatible API Changes](https://github.com/NREL/SAM/wiki/Checking-for-PySAM-Incompatible-API-Changes).

## Checklist:


cpaulgilman (commented):
The functionality looks good.

When I choose the option to design the tidal converter, I see both the tidal design inputs and the inputs for the “choose tidal converter from library” option. The power curve table is technically disabled, but it is so prominent that I think most users will not know to ignore it, in spite of the red text at the top of the page. I think this should be set up like a battery dispatch page where the radio buttons are in a separate top panel and control what appears in the bottom panel.

mjprilliman:
> The functionality looks good.
> 
> When I choose the option to design the tidal converter, I see both the tidal design inputs and the inputs for the “choose tidal converter from library” option. The power curve table is technically disabled, but it is so prominent that I think most users will not know to ignore it, in spite of the red text at the top of the page. I think this should be set up like a battery dispatch page where the radio buttons are in a separate top panel and control what appears in the bottom panel.

I can't replicate this. When I switch to the tidal turbine design I only see the design page.

mjprilliman:
@cpaulgilman Are you testing this on a Mac? I'm noticing the mac and Linux tests failing on SAM here (although only the pull_request) and I wanted to make sure there's not any differences across platforms. Also are you checking out the same branch name in ssc (https://github.com/NREL/ssc/pull/1181)?

brtietz:
@mjprilliman The distinguishing feature seems to be that the pull request builds are failing and the push builds are passing. The pull request builds are building against patch of ssc:

image

which seems incorrect in this case. I think it's safe to ignore it in this case and @dguittet and I can double check the CI scripts to fix this in the future.

mjprilliman:
> @mjprilliman The distinguishing feature seems to be that the pull request builds are failing and the push builds are passing. The pull request builds are building against patch of ssc:
> 
> image
> 
> which seems incorrect in this case. I think it's safe to ignore it in this case and @dguittet and I can double check the CI scripts to fix this in the future.

Thanks. I was looking for this yesterday but was looking in the wrong place. I had to rebase these branches to patch (manually, not using the git rebase command) if that is helpful in troubleshooting the CI.

dguittet:
Ok I think the issue is that the PR workflow should also use the related ssc branch if it's available. I'll make that change.

cpaulgilman (approved):
Looks good. In my previous review, I did not link this with https://github.com/NREL/ssc/pull/1181.

This requires a documentation update, which I will note for the Fall 2024 release.

PR #1810, opened 2024-07-18
Patch to develop 240718

brtietz:
Pairs with https://github.com/NREL/ssc/pull/1184 Tests may fail due to branch naming

For review: please confirm merge of equations.cpp went as intended.


src/equations.cpp
```diff
@@ -220,7 +212,8 @@ bool EqnDatabase::PreProcessScript( wxString *text, wxArrayString* errors)
 					arg[1] = compute module name
 					arg[2] = sim_type value
```
sjanzou, line 213: Looks correct based on issue 1634.

sjanzou (approved):
equations.cpp appears correct based on issue 1634.

@brtietz ,  should we merge ssc PR 1184 first and then update the ci.yml to use the develop branches and update the test results on Windows using develop branch of ssc and this branch of SAM and commit and push and check GitHub Actions?

brtietz:
> equations.cpp appears correct based on issue 1634.
> 
> @brtietz , should we merge ssc PR 1184 first and then update the ci.yml to use the develop branches and update the test results on Windows using develop branch of ssc and this branch of SAM and commit and push and check GitHub Actions?

https://github.com/NREL/SAM/issues/1813 resulted from attempting to do that. I'm open to merging this sooner (since develop will be less broken than it is with ssc merged) but also ok to hold off for a few days while I fix 1813.

PR #1827, opened 2024-07-31
Fix ci for PR vs push

dguittet:
- Want to match up the SSC branch of the same name, if that exists
- for Push, there's no destination branch, so the default branch needs to be stated explicitly in ci.yml
- for PR, there's a destination branch, so use that as the default branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,15 +206,16 @@ jobs:
         SAMNTDIR=$GITHUB_WORKSPACE/sam
         echo "SAMNTDIR=$SAMNTDIR" >> $GITHUB_ENV
         
-    - name: Get branch name to match with other repos
+    - name: Get branch name to match with other repos for push
+      if: ${{ github.event_name == 'push' }}
       run:  |
-          echo "GIT_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
+        echo "GIT_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
     
-    - name: Get base branch name for pull request
+    - name: Get branch name to match with other repos for pull_request
       if: ${{ github.event_name == 'pull_request' }}
       run:  |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -57,15 +57,15 @@ jobs:
...
 
-    - name: Get branch name for pull request
-      if: ${{ github.event_name == 'pull_request' }}
+    - name: Get branch name to match with other repos
       run:  |
-          echo "GIT_BRANCH=$GITHUB_BASE_REF" >> $GITHUB_ENV
-
-    - name: Get branch name for push
-      if: ${{ github.event_name == 'push' }}
+          echo "GIT_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
```
brtietz, line 62: This did not have the intended effect.

image

SSC won't have an 1827/merge ref, so this will always default to patch. I think this does return the correct value for a push, but not a pull_request.

Is the right move to override with `github.head_ref` in the if block below?

.github/workflows/ci.yml
```diff
@@ -57,16 +57,17 @@ jobs:
...
-
-    - name: Get branch name for push
+    - name: Get branch name to match with other repos for push
       if: ${{ github.event_name == 'push' }}
       run:  |
-        echo "GIT_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
+          echo "GIT_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
     
+    - name: Get branch name to match with other repos for pull_request
+      if: ${{ github.event_name == 'pull_request' }}
```
brtietz, line 66: This works for the Ubuntu pull request build. It looks like mac needs the same update:

image

dguittet: Yes updated, sorry for the slow step by step on this PR.

brtietz (approved):
Latest version looks good, thank you!

Issue #1830, opened 2024-08-02
REopt API call fails with callback error


cpaulgilman:
### Describe the bug

REopt API call fails.

### To Reproduce

1. Create a default PV Battery / Commercial case.
2. On Battery Cell and System page, click **Get size and dispatch**.
3. See error.

```
Could not evaluate callback function:call_reopt->on_change
[71]: error in call to 'reopt_size_battery()': access violation: expected hash table, but found null[12]: eval error in statement list
```

### Operating System and Version

 - macOS Sonoma 14.6
 - SAM 2023.12.17 r2 release candidate]

cpaulgilman:
As of the latest Patch branch (8/12/2024), we are seeing this error on macOS, but not Windows or Linux.

brtietz:
Closing since we merged PRs related to this for the patch. @sjanzou and @cpaulgilman please re-open if I've forgotten something that needs to be resolved on our end for this issue.

sjanzou:
> Closing since we merged PRs related to this for the patch. @sjanzou and @cpaulgilman please re-open if I've forgotten something that needs to be resolved on our end for this issue.

I thinks everything works fine with postman but, for curl, the issue remains and the pull request does have the REopt call for loads if the original call fails. 

Finally, the REopt team has been made aware of the problem and SAM will work with any non-API breaking fixes that are donw by the REopt team for curl calls.

PR #1836, opened 2024-08-08
Addresses issue #1830

sjanzou:
@cpaulgilman has completed all the work.

Update fcall_reopt_size_battery to dismiss simulation dialog and responsive reopt sizing progress dialog

Issue was wxWidget update to wxProgressDialog was blocking simulation dialog destruction and closing the reopt api dialog before the api call to v3 had completed.

Note that the PV-battery / Commercial default case resets  progress bar from 99% to 98% about 5 times.
Note that the PV-battery / Residential default case completes with the progress bar at about 33%.

Details in attached Word document.
[REopt2024.8.7_SAM_issue_1830.docx.zip](https://github.com/user-attachments/files/16541466/REopt2024.8.7_SAM_issue_1830.docx.zip)


cpaulgilman (approved):
Thanks for finding and fixing the issue with progress bars.

I increased the delay between REopt API status checks to account for long optimization run times for Default PV-Battery / Commercial case.

PR #1842, opened 2024-08-10
Help revision MPST Tower and Receiver total piping loss equation

cpaulgilman:
Final help revision for 2023.12.17 Patch 2


sjanzou (approved):
Thanks for updating - merging into patch for 2023.12.17.r2

PR #1844, opened 2024-08-15
Fix ci for PR vs push

dguittet:
Same as #1827 but for develop


brtietz:
@dguittet It looks like this might have been taken care of by bringing the changes over from patch in https://github.com/NREL/SAM/pull/1847. Can you confirm whether or not this PR is needed?

PR #1865, opened 2024-10-04
Standalone battery metrics

### LINKED ISSUE #1589: Set up battery charge from grid metrics for summary page of standalone battery
> Standalone battery capacity factor will always be infinity as currently calculated ("capacity_factor" isn't provided to cmod_battery.cpp, so nameplate_in is set to its initialized value of zero, leading to Inf. Should we take it out of the table or come up with a capacity factor for standalone battery?

1. I think we should remove capacity factor for Standalone battery configurations. 
2. It also seems like "Battery charge energy from system" will always be zero, so maybe that should be replaced by battery charge energy from grid, if there is such a metric.

_Originally posted by @cpaulgilman in https://github.com/NREL/SAM/issues/1561#issuecomment-1815573904_

mjprilliman:
## Description

-Add year 1 standalone battery metrics to summary table

Fixes #1589 

## Type of change

## Checklist:


cpaulgilman (approved):
Looks good. Thanks for adding the metrics.

PR #1900, opened 2024-11-07
IPH models pull request

sjanzou:
Please test and resolve issues to merge into for release here!
Goes with https://github.com/NREL/ssc/pull/1238


janinefreeman:
@taylorbrown75 @tyneises Is it still called a PPA for a thermal agreement? That terminology is on the revenue page for single owner.

cpaulgilman:
Some initial comments.

For IPH / Commercial configurations:

* There is no Thermal Load page. Should there be one?

* Construction financing has no effect on cash flow.

Results page Summary tab:

* In "Monthly AC Energy and Load in Year 1" plot, "energy" is ambiguous.

* "Electricity Net Generation" in kWh doesn't make sense. (This is the same as Janine's comment above)

cpaulgilman:
@taylorbrown75 For IPH configurations with the Single Owner financial model, I think the "Electricty Rates" page name in the navigation menu should be "Electricity Purchases" to be consitent with front-of-meter CSP and other configurations where electricity purchases are an operating expense. This is to distinguish from behind-the-meter configurations where "electricity rates" are used to calculate electricity bill savings. (You can see that there are separate Help topics for "Electricity Rates" and "Electricity Purchases" so we have separate places to explain how those inputs work.)

taylorbrown75:
> @taylorbrown75 For IPH configurations with the Single Owner financial model, I think the "Electricty Rates" page name in the navigation menu should be "Electricity Purchases" to be consitent with front-of-meter CSP and other configurations where electricity purchases are an operating expense. This is to distinguish from behind-the-meter configurations where "electricity rates" are used to calculate electricity bill savings. (You can see that there are separate Help topics for "Electricity Rates" and "Electricity Purchases" so we have separate places to explain how those inputs work.)

Fixed with (https://github.com/NREL/SAM/pull/1900/commits/c06063bada737e7c8f7d16743fcb8a242dce2e37)

Issue #1906, opened 2024-11-12
Adjust Losses UI forms list order is backwards from IDE


brtietz:
### Describe the bug

Adjust losses that are set to annual in the UI show up as subhourly in practice. It's a mismatch between widgets.h:
![image](https://github.com/user-attachments/assets/671e2efc-112e-4765-8d4b-fb2e6df070d7)

and UIObjects.cpp

![image](https://github.com/user-attachments/assets/ef153ccb-e685-4c84-b9d0-703716cd7ef5)

These lists should be in the same order so that UI IDE users get what they expect

It'll be easier to reorder widgets.h and maintain the existing selections in the IDE

PR #1907, opened 2024-11-12
Fix order issue #1906 and do not allow free form combo box inputs

sjanzou:
Please test the Detailed PV DC and AC losses and any use of the wxUILossAdjustmentCtrl and anywhere the AFDataLifetimeMatrixDialog and AFDataLifetimeArrayDialog are used - e.g. AFDataLifetimeMatrixDialog


brtietz (approved):
Thanks for updating this! This corrected the behavior such that it's driven by default values rather than anything else. As per widgets.cpp 2635 any value that isn't easily divisible by the analysis period defaults to subhourly.

I've added some suggested defaults changes to get everything on single value, but I'm open to other approaches (such as annual to match the mode in the UI form).

mjprilliman (approved):

cpaulgilman (approved):

PR #1908, opened 2024-11-12
Cmake edits

dguittet:
Fixes https://github.com/NREL/ssc/issues/1118 along with https://github.com/NREL/ssc/pull/1243


dguittet:
Not sure why the tests are failing now with large changes to a specific SAM case, but I'll take a look

image

dguittet:
Not sure why but retrying the test fixed it.. Wonder if there's some issue with the Generic CSP System Leveraged Partnership Flip case being indeterminate...?

Issue #1903, opened 2024-11-08
Thermal Rates and Thermal Load pages and compute module missing from Hybrid fuel cell host developer configurations


sjanzou:
The Hybrid host developer configurations should include Thermal Rates and Thermal Load page similarly to the Fuel Cell Commercial configuration
![image](https://github.com/user-attachments/assets/7714756d-29e1-43cb-b930-90c82f078ed1)

And the cmod_thermalrate compute module should be called after the cmod_utilityrate5 compute module is called
![image](https://github.com/user-attachments/assets/8af9bb39-f2d1-4b54-a9a3-c297b997c8c9)

Currently, the hybrid fuel cell configurations do not include the Thermal Rates and Thermal Load page and do not call the thermalrate compute module:
![image](https://github.com/user-attachments/assets/a7bb5ba2-bf7e-4827-a489-7534e9781a99)

![image](https://github.com/user-attachments/assets/21419014-c801-48a7-8fca-206a4a3e391b)

PR #1911, opened 2024-11-15
Address SAM #1903 

sjanzou:
Goes with https://github.com/NREL/ssc/pull/1244

Please check proforma - added thermal value to Host returns and updated defaults.

Test file for testing with Fuel Cell Commercial and all Fuel Cell Hybrids with Host Developer

[SAM_1903.zip](https://github.com/user-attachments/files/17774590/SAM_1903.zip)


cpaulgilman (approved):
Thanks for fixing, and for adding thermal value rows to cash flow.

PR #1926, opened 2024-11-23
UI fixes 2024

cpaulgilman:
Check alignment, text box sizes, etc. on Linux for all UI updates for Fall 2024.

Go over new IPH input pages and set widget Group and UIObject property.

@dguittet I'm listing you as a reviewer since it looks like you're assigned to check UI forms on Mac -- please do that check from this branch or after it is merged into Develop.


cpaulgilman:
@sjanzou Thanks for finding those issues. I'm finding that I need to set the height for list boxes to 28 to make the text fit on Linux.  Other widgets seem to work well with height = 24. (Many currently have height = 21.)

I also did a spot check to make sure these changes for Linux don't make things look bad in Windows.

cpaulgilman:
@taylorbrown75 Is it OK to delete the "Electricity Purchases Heat Model.json" file and its [associated commented-out line in startup.lk](https://github.com/NREL/SAM/blob/5343436e23108518e9395aa63de739ec61903ff8/deploy/runtime/startup.lk#L1982)? It looks like this might have been left over from an earlier iteration.

sjanzou (approved):
Looks good on Windows and Linux. I will defer to @dguittet for MacOS UI review.

taylorbrown75:
> @taylorbrown75 Is it OK to delete the "Electricity Purchases Heat Model.json" file and its [associated commented-out line in startup.lk](https://github.com/NREL/SAM/blob/5343436e23108518e9395aa63de739ec61903ff8/deploy/runtime/startup.lk#L1982)? It looks like this might have been left over from an earlier iteration.

That should be okay to remove. Seems to be from before we updated the IPH financials.

PR #1931, opened 2024-11-28
GETEM Scenario updates based on emails from Erik 11/27/2024

sjanzou:


mjprilliman (commented):
I believe the contingency needs to be changed to 10% for all json cases. I can go through and update that.

sjanzou:
> I believe the contingency needs to be changed to 10% for all json cases. I can go through and update that.

Thanks for updating!

sjanzou (commented):
Thanks for updating the contingency

PR #1948, opened 2024-12-05
Quickstart, macro, and sample file upgrades

janinefreeman:
Updates the mini grid macro file to the latest SAM version (note to reviewers - please test that quickstart function), the PV subarray layout optimization macro for new bounds on the GCR input, and small updates to the sample scripts.


dguittet:
Works-- really nice example and plots!

I am getting this old SAM version notification though
image

dguittet:
The Subarray macro has tables that use GCR=1 that need to be updated with the new max value of 0.99:
image

Also, why do some rows have commas and others don't?

janinefreeman:
Hmm, that's super weird that you're getting the old file version upgrade notice for the mini-grid example- it shows as updated on my computer. I just tried re-saving it and pushing it, can you see if it works now? If not, could this be a Mac vs Windows issue?

janinefreeman:
Good thought about making analysis period 1 temporarily, that's MUCH faster. I didn't write the macro originally, just updating it to work with the latest version, so not sure why the number formatting is different in different rows... But the table shows the correct max GCR for me (everything in the table is keyed off of the max GCR input). Could it be that your local build didn't sync correctly with this one?
temporary

dguittet:
I'm not getting the version upgrade notice anymore!

But I'm still getting the extended table for the array macro, even after manually copying over the new code. But the table looks the same in the 2023.12.17 SAM, so might be a Mac thing...

dguittet:
I tested it on Windows and it's the same. What I did was run the array macro on the file created with the Quick Start for Puerto Rico.

mjprilliman (approved):
Mini grid wizard works on Windows

Issue #1918, opened 2024-11-20
Code Generator providing incorrect sscapi.h - Matlab code generation fails


sjanzou:
See discussion at https://sam.nrel.gov/forum/forum-sdk/4788-matlab-and-ssc-library.html

sjanzou:
To fix, include the latest sscapi.h file into the x64 folder with the ssc.dll file in the distribution and then MATLAB will find the exported function.

For completeness, please add sscall('unload'); to the beginning of the generated Matlab code to ensure the correct ssc.dll is loaded; e.g.

clear
ssccall('unload');
ssccall('load');

PR #1950, opened 2024-12-06
Fix SAM #1918 and test on Windows 11 and MacOS 15.1.1 with MATLAB

sjanzou:
To test, build SAM and SAM-private if you would like (delete build_windows folder in SAM-private) and run SAM and generate code for some language and check that latest sscapi.h and ssc.dll (or.so or .dylib) is included in the generated code folder.


sjanzou:
Tested on widows 11
image


Tested on MacOS 15.1.1
image

src/codegenerator.cpp
```diff
@@ -193,12 +193,13 @@ bool CodeGen_Base::PlatformFiles()
...
 	// sscapi.h - switch to copy version for syching issues 5/30/17
 	// assumes in runtime folder for all builds.
-	f1 = SamApp::GetAppPath() + "/sscapi.h";
+	f1 = SamApp::GetRuntimePath() + "/sscapi.h";
 	if (wxFileExists(f1)) {
 		f2 = m_folder + "/sscapi.h";
 		wxCopyFile(f1,f2);
 	}
-	else {
+	//else {
```
mjprilliman, line 201: Is this going to be needed in the future? If not, can we remove the commented part?

src/codegenerator.cpp
```diff
@@ -193,12 +193,13 @@ bool CodeGen_Base::PlatformFiles()
...
+		/* skip fallback per SAM issue 1918
```
mjprilliman, line 202: Same question

sjanzou: Removed commented code as requested - was initially left as a fallback position but more confusing and out of date. Thanks for the future maintenance request!

PR #1955, opened 2024-12-06
Update test results for me cost model changes

mjprilliman:
-Update results for changes in https://github.com/NREL/ssc/pull/1249


mjprilliman:
Closed as changes are already included in https://github.com/NREL/SAM/pull/1958

sjanzou:
> Closed as changes are already included in #1958

We should come up with a better process of updating the default results and the api files; especially when there is a flurry of pull requests and failing GitHub Actions...

PR #1967, opened 2024-12-10
Update case comparison macro

mjprilliman:
## Description

-Exit from macro for Third Party - Host, as there are no financials
-Fix after-tax cash flow variable for Host Developer configs
-Fix typo in error message for non-supported financial models


cpaulgilman (approved):
Looks good.

For future consideration: For battery cases, it might be useful to show battery roundtrip efficiency. Note that capacity factor for Standalone Battery configs is reported as infinity (untitled (2) case in this screenshot):

image

PR #1988, opened 2025-01-03
add getem eqn to SAM_eqns

dguittet:
Adds `SAM_getem_om_cost_calc_eqn` so it can be called from PySAM

Note the updated SAM api code from export_config using the latest SSC getem_costs branch


sjanzou (approved):
Updated include file in SAM_eqns.cpp 
PySAM built and working with access to getem_om_cost_calc
image

PR #1999, opened 2025-01-30
Fixes #1965

### LINKED ISSUE #1965: Revenue "apply values from library" function doesn't work for hybrids
### Describe the bug

The PPA revenue has a library of time of delivery values that can be applied to the current table using a button. This works for non-hybrid models but does not work in hybrids.

### To Reproduce

1. Open a hybrids case
2. Go to the revenue page
3. Select a different PPA TOD schedule (such as generic duck curve)
4. Attempt to apply the values
5. See error

### Expected behavior

Values should be applied, same as the other models

 - Version 2023.12.17, 2024.12.12 release candidate

sjanzou:
Please test with hybrid and non-hybrid configurations

### DIFF
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -1233,6 +1233,7 @@ void fcall_value( lk::invoke_t &cxt )
 	wxString name = cxt.arg(0).as_string();
 	auto& c = cc.GetCase();
 	bool found = false;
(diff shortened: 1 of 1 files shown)


brtietz (approved):
This fixed it, thank you!

PR #2002, opened 2025-02-05
Add steam heat sink for IPH models

tyneises:
This pull request corresponds to SSC pull request 1274 https://github.com/NREL/ssc/pull/1274.

It adds an option to indirect IPH technologies to model a steam heat sink that is heated by the field HTF through a counterflow heat exchanger.


taylorbrown75 (approved):
Tested on my machine, works well.

PR #2012, opened 2025-02-18
Combine cases macro checks lifetime mode

### LINKED ISSUE #1752: Combine Cases Macro doesn't check simulation timesteps
### Describe the bug

The combine cases macro will gleefully combine cases with different weather file timesteps, leading to odd phenomena like PV generating at night.

### To Reproduce

1. Set up a SAM case with two PV cases and a generic system
2. Give one PV case an hourly weather file, and another case a subhourly file
3. Set up a generic system to combine the cases, and run
4. See PV generating at night!

### Expected behavior

The combine cases macro should at least throw an error if a user tries to do this, or maybe if we have lots of time interpolate to a consistent timestep.

brtietz:
# Pull Request Template

## Description

Add code to check whether or not a simulation uses lifetime mode prior to assessing whether or not the simulation is subhourly

Fixes #1752

To test: Create a PVWatts case with a subhourly weather file, a PVWatts case with an hourly weather file, and a custom generation case that combines these two cases. The combine cases script should flag the fact that a case is subhourly and give an error that it is not currently supported.

## Type of change

### DIFF
--- a/src/combinecases.cpp
+++ b/src/combinecases.cpp
@@ -262,7 +262,9 @@ void CombineCasesDialog::OnEvt(wxCommandEvent& e)
 							analysis_period_this = current_case->Values(0).Get("c_lifetime")->Value();
 						}
 						else if (financial_name != "None") {
+							bool use_lifetime = current_case->Values(0).Get("system_use_lifetime_output")->Boolean();
 							analysis_period_this = current_case->Values(0).Get("analysis_period")->Value();
+							analysis_period_this = use_lifetime ? analysis_period_this : 1;
 						}
 
 						// determine hourly generation profile of current case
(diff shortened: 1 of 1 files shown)


sjanzou (changes requested):
Can you please change the error message from the technology name to the case name? For example in the attached project file, change "PVWatts" to "subhourly" in the error message:
image

Referenced project file
[SAM_PR_2012.zip](https://github.com/user-attachments/files/18862673/SAM_PR_2012.zip)

brtietz:
> Can you please change the error message from the technology name to the case name? For example in the attached project file, change "PVWatts" to "subhourly" in the error message: image
> 
> Referenced project file [SAM_PR_2012.zip](https://github.com/user-attachments/files/18862673/SAM_PR_2012.zip)

Thanks for testing! Fixed in latest push.

sjanzou (approved):
Great! Thanks for addressing the subhourly/hourly combination!

PR #2017, opened 2025-02-21
Patch to develop 2/21/25

brtietz:
Merge the last few weeks of patch PRs into develop


brtietz:
GH Actions pulled the wrong SSC branch. Expectation is that issue will be fixed post-merge.

PR #2035, opened 2025-03-13
Remove SAM API files based on removed ssc cmods. 

brtietz:
Some defaults changes in export config got captured here too


sjanzou (approved):
Thanks for updating SiingleownerHeat_DSGLIPGSingleOwner default api

Issue #1426, opened 2023-08-06
Create an error if someone enters Lat/Long pairs with N,s,e,w and show them the proper format.


sjanzou:
![image](https://github.com/NREL/SAM/assets/6498311/baf6b5c2-770a-4390-a4ca-19305dc3e2a4)

Parse user entered data more completely and do not return invalid date

cpaulgilman:
This needs to be handled consistently for solar, wind, and wave resource data downloads.

My suggestion is that we use [decimal degrees](https://en.wikipedia.org/wiki/Decimal_degrees) as the standard way to input geographic coordinates in SAM, and return an error if degree (°), minute ('), or second (") symbols or "N/S/E/W" are included in the request.

The Solar Resource Data UI callback attempts to identify non-decimal-degree requests, but it is probably best to do this in the geocode() function instead: https://github.com/NREL/SAM/blob/58e2e7f44d2ebf23a308cf1e967c75c192e8694b/deploy/runtime/ui/Solar%20Resource%20Data.txt#L2735.

There are two Geocode functions. One uses the Google API for non-NREL builds:

https://github.com/NREL/SAM/blob/58e2e7f44d2ebf23a308cf1e967c75c192e8694b/src/geotools.cpp#L51

The other uses an NREL Developer API for NREL builds:

https://github.com/NREL/SAM/blob/58e2e7f44d2ebf23a308cf1e967c75c192e8694b/src/geotools.cpp#L112

Here are some examples of strings people might use for geographic coordinates:

45° 53' 8.1" N 8° 57' 29.9" E

45° 53' 8.1" N, 8° 57' 29.9" E

45°53'8.1"N,8°57'29.9"E

45°53' N 8°57' E

45°53' N, 8°57' E

45°53' N 8°57' E

45°53'N,8°57'E

45°53' N 8°57' E

40° 42′ 51″ N 74° 0′ 21″ N 40 W

15°46'47" S 47°55.783' W

15.7797° S 47.9297° W

cpaulgilman:
Python Lat/Lon parser from NOAA: https://github.com/NOAA-ORR-ERD/lat_lon_parser/tree/master

cpaulgilman:
LK script:

```
x = '45° 53\' 8.1\" N 8° 57\' 29.9\" E';
outln(x);

y = geocode(x);
outln(y);
```

![image](https://github.com/user-attachments/assets/f78688af-e187-4a07-8187-7764f663aecf)

PR #2039, opened 2025-03-21
Fix geocode crash

### LINKED ISSUE #2036: SAM crashes when invalid address requested for NSRDB download
### Describe the bug

If you enter an invalid location name or address for the NSRDB weather file download, SAM crashes. It should display an error message instead.

### To Reproduce

1. Create a default PVWatts case.
2. On the Location and Resource page, under "Download Weather Files" enter "xxxxxxxx" for the location name.
3. Click **Download and add to library**
4. SAM crashes.

 - Windows 11
 - SAM 2024.12.2

cpaulgilman:
Fixes crash caused by submitting an invalid address to geocode.

* JSON check was invalid when results field is empty.

* Don't initialize lat/lon/tz to zero because these are valid values.

Fixes #2036

To test (requires building SAM-private or modifying private.h):

1. Create a default PVWatts case.
2. On the Location and Resource page, under "Download Weather Files" enter "xxxxxxxx" for the location name.
3. Click Download and add to library
4. SAM should display a message instead of crashing.

### Checklist
- [ ] changes defaults

### DIFF
--- a/src/geotools.cpp
+++ b/src/geotools.cpp
@@ -51,6 +51,7 @@ static wxString MyGet(const wxString& url)
 }
 */
 // Geocode using Google API for non-NREL builds of SAM
+// Set key in private.h: Google API key
 bool GeoTools::GeocodeGoogle(const wxString& address, double* lat, double* lon, double* tz, bool showprogress) {
     wxBusyCursor _curs;
 
@@ -171,6 +172,8 @@ bool GeoTools::GeocodeGoogle(const wxString& address, double* lat, double* lon, 
 }
 
 // Geocode using NREL Developer API (MapQuest) for NREL builds of SAM
+// Set keys in private.h: NREL Developer API key for geocode, and Bing API key for time zone
+// Set variables in webapis.conf to URLs: bing_maps_timezone_api, nrel_geocode_api
 bool GeoTools::GeocodeDeveloper(const wxString& address, double* lat, double* lon, double* tz, bool showprogress) {
     wxBusyCursor _curs;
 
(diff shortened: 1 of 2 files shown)


sjanzou (commented):
In the sam-2036-invalid-address-crash branch of SAM, the "xxxxxxx" location always downloads 38.8904_-77.032 as shown by the log file image here:

image

Same as for current release
image

Not sure if the API changed for the geocoding????


@brtietz , @janinefreeman are your receiving the same results with the 2024.12.12 release and the build from the sam-2036-invalid-address-crash branch?

brtietz:
> @brtietz , @janinefreeman are your receiving the same results with the 2024.12.12 release and the build from the sam-2036-invalid-address-crash branch?

Same results on my machine. I agree this seems like the API changed. Is it worth trying to reproduce the crashing responses with a mock server? https://learning.postman.com/docs/design-apis/mock-apis/set-up-mock-servers/

cpaulgilman:
> @brtietz , @janinefreeman are your receiving the same results with the 2024.12.12 release and the build from the sam-2036-invalid-address-crash branch?
>
> Same results on my machine. I agree this seems like the API changed. Is it worth trying to reproduce the crashing responses with a mock server? https://learning.postman.com/docs/design-apis/mock-apis/set-up-mock-servers/

Thanks for testing. I agree this seems like it may have been an intermittent issue with the API. The mock server looks useful for testing. The crash was caused by the API returning different JSON than SAM was expecting. SAM should be robust enough to handle that without crashing. I'll investigate further.

cpaulgilman:
More robust handling of different responses from the Geocode API will be handled in the fix for #1426 with test responses from Postman mock servers. Closing this here.

PR #2043, opened 2025-03-28
Marine UI updates

mjprilliman:
### Description
-Update ME cost curves
-Fix modeled cost variables assigned to GUI equations rather than cmod_mhk_costs ouptuts
-Closes #1974 

To review cost curves: 
-See email with cost curve information

To review JPD file generation:
-Open ME Wave -LCOE calculator
-Click 'Create JPD from time series weather file(s)' button
-Use attached folder as weather files (or download weather files from API on SAM-private)
-Save output jpd
-Switch from 'Time series' to 'Frequency of occurence', check for output jpd file without callback errors

[test_jpd.zip](https://github.com/user-attachments/files/19507853/test_jpd.zip)

Fixes #1974 

### Corresponding branches and PRs:
marine-cost-branches in ssc

SSC PR: https://github.com/NREL/ssc/pull/1305

### Unit Test Impact:

-Need to re-run test script

### Checklist
- [ ] requires help revision and I added that label
- [ ] adds, removes, modifies, or deletes variables in existing compute modules
- [ ] adds a new compute module
- [ ] changes defaults
- [ ] I've tagged this PR to a milestone

### _Reminders- this section can be deleted_
[Checking for PySAM Incompatible API Changes](https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files).

[When do the PySAM files need to be regenerated?](https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files#when-do-the-pysam-files-need-to-be-regenerated-via-export_config)

PR #2047, opened 2025-04-01
Update JSON loading and saving to handle NaN and Inf 

### LINKED ISSUE #1856: Allow "inf" as a GUI input options for utility tiers
See https://github.com/NREL/SAM/issues/1834 for background

1e38 was based on 32-bit SAM. Now that we're 64 bit max double is actually larger. Implementing this as inf (infinity) in the GUI is potentially clearer to users and gives us flexibility on the backed on how big of a number to represent.

sjanzou:
Addresses SAM issue #1856 to use Inf values as tier maximum usages.

Fixes #1856

Goes with ssc pull request https://github.com/NREL/ssc/pull/1309

expected changes in test_results files - doubles are written out to full precision and NaN and Inf are directly supported - expect results to differ slightly.

### Checklist
- [ x] changes defaults
- [x ] I've tagged this PR to a milestone

PySAM regeneration should be tested 

ME Tidal LCOE calculator defaults seems to have an issue in the equations.


mjprilliman (commented):
image

Energy charges with max usage of 'nan' gives the same results as max usage 'inf'. What does max usage 'nan' represent? Not a super likely use case, just curious 

Not sure where this would be relevant, but this also allows for -nan and -inf inputs. 

https://github.com/NREL/SAM/pull/2043 needs to be merged to fix the ME tidal issue

sjanzou:
> image
> 
> Energy charges with max usage of 'nan' gives the same results as max usage 'inf'. What does max usage 'nan' represent? Not a super likely use case, just curious
> 
> Not sure where this would be relevant, but this also allows for -nan and -inf inputs.
> 
> #2043 needs to be merged to fix the ME tidal issue

Thanks for testing. The marine-ui-updates branch was merged in to fix the defaults issue.

ssc pull request https://github.com/NREL/ssc/pull/1309 was created to address the nan, -nan, and -inf handling.

Now, 1e+38 and inf for max tiers gives same result and negative values and NaNs throw errors
image

image

cpaulgilman:
This specifically addresses #1856 by adding support for NaN and Inf in the DataMatrix widgets for electricity rate inputs. However, it is not possible to type "Inf" for numeric inputs with default value of 1e38 like incentives. Do we want NaN and Inf to work for all numeric inputs?

image

sjanzou:
> This specifically addresses #1856 by adding support for NaN and Inf in the DataMatrix widgets for electricity rate inputs. However, it is not possible to type "Inf" for numeric inputs with default value of 1e38 like incentives. Do we want NaN and Inf to work for all numeric inputs?
> 
> image

@cpaulgilman ,  that was the next question (or issue) to replace all 1e+38 with inf in defaults. We now support saving and loading inf and NaN values directly in JSON.

PR #2063, opened 2025-04-10
Make stimulation cost per well calculated, update api

mjprilliman:
### Description
-Stimulation cost per well box needs to be calculated not input
-Update API from testing PySAM (getem_om_cost_calc works as intended)

### Checklist
- [ ] changes defaults


sjanzou (approved):
Many api updates...

PR #2079, opened 2025-04-22
PV Snow Macro: Fix conversion when snow data expressed in inches

berg-michael:
### Description

If one attempts to use the macro available in various PV configurations to append snow data measured in inches, the conversion step of the macro will fail with an unintuitive error message. This PR fixes this bug. 

To test:
Create a new PV system. Download a weather file. Download the attached snow data. Attempt to append it to the weather file you download, selecting a header of size 1, a column name of (Observed) Snow Depth (in), and units as inches. The macro will fail. This PR fixes that failure. 

Fixes # (issue(s)): None yet. 

None applicable. 

### Unit Test Impact:

None applicable.

### Checklist
- [ ] changes defaults

[denver, co_DVNC2_observed_snow_depth.csv](https://github.com/user-attachments/files/19858391/denver.co_DVNC2_observed_snow_depth.csv)

### DIFF
--- a/deploy/runtime/macros/Flat Plate PV/Append Snow Data.lk
+++ b/deploy/runtime/macros/Flat Plate PV/Append Snow Data.lk
@@ -128,7 +128,7 @@ if( mod(weatherLength, 8760) != 0){
 // Load snow data
 
 if( sf_unit == 'cm' ) conversion = 1.0;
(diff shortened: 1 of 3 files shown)


cpaulgilman (approved):
Looks good. This fixes the snow depth conversion from inches to cm.

PR #2086, opened 2025-06-03
Fix defaults manager handling of json files with NaNs. Fixes #2085

### LINKED ISSUE #2085: Defaults manager does not handle NaN values
There is an issue with the handling of NaN values in JSON. When I attempt to add a new variable (with non-NaN value), I get a diff that basically truncates the defaults after the first NaN in the original file. See the attached picture for an example diff.

 

To reproduce this behavior:

Open SAMOS
Open Defaults Manager
Select “Flat Plate PV,None” in the left sidebar
In “Variable Name”, input use_snow_weather_file
Click query and observe this variable does not exist
Check “Change data type” and “Add variable”.
Set “Data type” to Number and “Enter Numeric Value” to 1
Click Modify
Observe that the deploy/runtime/defaults/Flat Plate PV_None.json is now truncated at line 373, which is the position of the first NaN value.

Image

sjanzou:
Fix #2085

### DIFF
--- a/src/case.cpp
+++ b/src/case.cpp
@@ -613,7 +613,6 @@ bool Case::SaveDefaults(bool quiet)
 
 
     rapidjson::StringBuffer os;
-//	rapidjson::PrettyWriter<rapidjson::StringBuffer> writer(os); // MSPT/MP 64MB JSON, 6.7MB txt, JSON Zip 242kB
 	rapidjson::PrettyWriter<rapidjson::StringBuffer, rapidjson::UTF8<char>, rapidjson::UTF8<char>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag> writer(os); // MSPT/MP 64MB JSON, 6.7MB txt, JSON Zip 242kB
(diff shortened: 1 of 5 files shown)


berg-michael (commented):
These changes look good to me and align with those made in SAM #1856. 

There are a few other places where we use PrettyWriter. Most have the kWriteNanAndInfFlag set, but  `InputPageData::Write_JSON` in `src/main.cpp` and` WritePythonConfig` in `src/pythonhandler.cpp` do not. 

I'm not familiar with the type of data that'd be written by those functions. Is is possible that there would be NaNs/Infs? If so, I think those should be changed as well. 

Otherwise, I have no comments.

sjanzou:
Go through remaining occurrences of PrettyWriter and update all  with NaN and Inf flag and create ssc pull request https://github.com/NREL/ssc/pull/1333 to go with SAM #2086

berg-michael (approved):
Looks good to me!

brtietz (approved):

Issue #2008, opened 2025-02-12
SAM crashes if "save all output variables over analysis period" is unchecked and a grid outage is specified


brtietz:
### Describe the bug

If a user unchecks the "save all output variables over analysis period" and tries to run a grid outage simulation, SAM crashes.

### To Reproduce

1. Set up a behind the meter, detailed PV + battery simulation (tested with host developer)
2. On the "DC degradation" page, uncheck the "save all output variables over analysis period" box
3. Set up a grid outage (tested with 100% crit load and an outage in every timestep
4. Run the simulation
5. SAM crashes

### Expected behavior

SAM should not crash in this scenario

### Operating System and Version

 - Version 2023.12.17, 2024.12.12

berg-michael:
My prior analysis of this bug. Should be fixed in SAM/SAM_2008 and ssc/SAM_2008 (PRs linked by Github above). 

> It's a heap buffer overflow, which is why the debugger was getting extremely confused.
>  
> The issue is in the following region (line 3103 of cmod_pvsamv1.cpp):
>  
>  
> 
> ```
> if (batt->is_outage_step(idx % 8760)) {
>     offline = batt->is_offline(idx);
> }
> if (offline && PVSystem->p_systemACPower[idx] < tolerance) {
>     PVSystem->p_inverterTotalLoss[idx] = (ssc_number_t)(PVSystem->p_inverterTotalLoss[idx] - PVSystem->p_inverterNightTimeLoss[idx]);   //numberOfWeatherFileRecords
>     ssc_number_t avoided_losses = PVSystem->p_inverterNightTimeLoss[idx] + PVSystem->p_acWiringLoss[idx] + PVSystem->p_transmissionLoss[idx];
>     batt->outGenWithoutBattery[idx] += avoided_losses; //nrec * nyears
>     batt->outUnmetLosses[idx] -= avoided_losses;        //nrec * nyears
>     PVSystem->p_inverterNightTimeLoss[idx] = 0.0; //numberOfWeatherFileRecords
>     PVSystem->p_acWiringLoss[idx] = 0.0;    //numberOfWeatherFileRecords
>     PVSystem->p_transmissionLoss[idx] = 0.0; //numberOfWeatherFileRecords
> }
> ```
>  
>  
> If there is an outage, then offline will be set, and we will enter the if statement. 
> If "Save all output variables over analysis period" is unchecked, then saveLifetimeVars will be set to zero in lib_pv_io_manager.cpp. 
[...]

PR #2087, opened 2025-06-30
UI: Disable option to not save lifetime variables for PV-Battery systems. 

berg-michael:
### Description

[SAM#2008](https://github.com/NREL/sam/issues/2008) describes a bug in which deselecting the option to save lifetime variables in a PV-Battery simulation leads SAM to crash. The underlying cause of the crash is described more thoroughly in the SAM issue, but briefly, it is a heap corruption bug caused by the smaller, non-lifetime allocations being too small to compute figures for the battery in cmod_pvsamv1.cpp.

This fixes this issue on the UI side by preventing users from deselecting the option to save lifetime variables for PV-Battery systems. It also provides a brief warning why this option is disabled. 

Fixes # (issue(s))
#2008 

[ssc/SAM_2008](https://github.com/NREL/ssc/pull/1334)

### Unit Test Impact:

Tests shouldn't be impacted. 

### Checklist
- [ ] changes defaults

### DIFF
--- a/deploy/runtime/startup.lk
+++ b/deploy/runtime/startup.lk
@@ -121,7 +121,7 @@ configopt( 'Community Solar', { 'long_name'='Community Solar', 'short_name'='Com
 configopt( 'None', { 'long_name'='No Financial Model', 'short_name'='No financial', 'description'='Run the performance model with no financial model' } );
 
 enum { LOAD_SIMPLE, LOAD_BELPE, NO_LOAD };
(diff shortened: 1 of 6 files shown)


brtietz (approved):
This works, thanks for fixing! I expected an issue with Detailed-PV Battery Wind in the hybrid models, but because that's AC connected it doesn't trip the code of concern here.

cpaulgilman (approved):
Looks good. I made some changes to the widgets for consistency. Thanks!

PR #2088, opened 2025-06-30
Parametrics: Ensure initialization before attempting to read or copy runs

### LINKED ISSUE #2013: Parametrics "create new case" causes SAM to crash
### Describe the bug

Creating a case from parametric runs (right-click row header in parametrics table) causes SAM to crash when there are no results in the parametrics table.

### To Reproduce

1. Open .sam file in [parametric-test.zip](https://github.com/user-attachments/files/18873374/parametric-test.zip).
2. Click **Parametrics**.
4. Right-click any row header and click *Create new case**.
5. SAM crashes.

SAM crashes consistently when you create a new case without first running parametrics.

### Expected behavior

When you try to create a new case from the parametrics table with no parametric results, SAM should either prompt you to first run parametrics, or it should create the new case without crashing. I don't think parametrics results should be required to create a new case.

On a related note, when SAM creates the new case, it includes a copy of the parametrics in the new case, which was confusing in the workflow I was testing. I expected the parametric setup to only be in the original case.

 - OS: Windows 11
 - Version: SAM 2024.12.12

berg-michael:
### Description

#2013 reports a bug in which creating new cases from parametric runs leads to a crash if the parametric runs haven't actually been simulated. 

@cpaulgilman 's steps to reproduce:

> Open .sam file in [parametric-test.zip](https://github.com/user-attachments/files/18873374/parametric-test.zip).
>     Click Parametrics.
>     Right-click any row header and click Create new case*.
>     SAM crashes.

I note that the crash also occurs if one chooses "Show inputs". The cause of this issue, and the solution, are the same as for the original reported issue. 

This bug is caused by attempting to read uninitialized data. 

This PR solves this problem by checking for initialization before this read, and if uninitialized, initializes. 

Fixes #2013 

This is a SAM-only change. 

### Unit Test Impact:

No unit test impact. 
### Checklist
- [ ] changes defaults

### DIFF
--- a/src/parametric.cpp
+++ b/src/parametric.cpp
@@ -852,6 +852,11 @@ void ParametricViewer::OnMenuItem(wxCommandEvent &evt)
 	case ID_SHOW_ALL_INPUTS:
 		if ((int)m_grid_data->GetRuns().size() > m_grid_data->GetRunNumberForRowNumber(m_selected_grid_row)){
 			if (m_grid_data->GetRuns()[m_grid_data->GetRunNumberForRowNumber(m_selected_grid_row)])	{
+				// If uninitialized, init and set up variables from the parametric grid
+				if (m_grid_data->GetRuns()[m_grid_data->GetRunNumberForRowNumber(m_selected_grid_row)]->GetInputSize() < 1)
+				{
+					m_grid_data->PrepareSimulations();
+				}
 				std::vector<VarTable*> pvts;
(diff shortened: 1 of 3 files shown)


cpaulgilman (approved):
Looks good to me. Thanks for the fix!

sjanzou (approved):
Thanks for the update and the details of all the bugs fixed in the pull request!

PR #2093, opened 2025-07-17
Migrate Bing Maps API to Azure Maps API

### LINKED ISSUE #2092: Weather file downloads with address cause SAM to crash
### Describe the bug

On Location and Resource page, typing a street address or location name and clicking **Download and add to library** causes SAM to crash.

Workaround solution is to use latitude/longitude pair instead of address. See https://sam.nrel.gov/forum/forum-general/5021-problem-downloading-weather-files.html#14154

### To Reproduce

1. Create a solar case.
2. Go to Location and Resource page.
3. Under "Download Weather Files" type a location name like "Portland, OR".
4. Clicking **Download and add to library**.
5. SAM crashes. 

Initial investigation suggests this is an issue with the time zone API call.

This also breaks the 3D Shade Calculator map loading function.

cpaulgilman:
### Description

Migrate [deprecated Bing Maps API](https://learn.microsoft.com/en-us/bingmaps/getting-started/bing-maps-dev-center-help/getting-a-bing-maps-key) to Azure Maps API. This pull requests modifies the following functions to use the Azure API, which currently do not work because we don't have an Azure API key for SAM:
   1. Time zone API for geocode function. 
   2. Static map API for 3D shade calculator map underlay
 
Make time zone optional for geocode() function to address (1) above.

Update 3D Shade Calculator to handle failed static map API calls for image underlays to address (2).

Update LK geocode() function to add optional parameter 

This is a temporary fix. For a more permanent fix:

1. Request update to NREL Developer Geocode API to return time zone. Or, consider an offline implementation such as one of the options listed in Stack Overflow [here ](https://stackoverflow.com/questions/16086962/how-to-get-a-time-zone-from-a-location-using-latitude-and-longitude-coordinates). Note that time zones do not correlate well with latitudes as can be seen in [this map](https://www.timeanddate.com/time/map/) from timeanddate.com. 

2. See if NREL Developer API can return a static map for 3D shade calculator. If not, investigate getting a Microsoft Azure key for SAM, or finding another static map service.

See https://github.com/NREL/SAM-private/pull/129 for more discussion of next steps.

Fixes #2092

Goes with https://github.com/NREL/SAM-private/pull/129

### To Test

1. Build SAM from sam-private.
2. Create a Detailed PV / Residential case.
3. On the Location and Resource page, try to download a weather file using a street address or location name. Download should work without crashing SAM.
4. On the Shading and Layout page, click **Open 3D shade calculator**,  go to the Location tab and try clicking **Lookup address** and **Update map from coordinates**. SAM should display error messages without crashing.
5. Click File, New script and run the following LK script:
```
// return lat, lon
x = geocode("portland, or");
outln(x);

// return lat, lon, tz=0 with error message
x = geocode("portland, or", true);
outln(x);
```

An optional test is to set the value of `azure_api_key` in `SAM-private/src/private.h` to a valid Microsoft Azure API key -- this should make Steps 4 and 5 work properly without error messages.

### Checklist
- [ ] changes defaults

### DIFF
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -3805,7 +3805,9 @@ void fcall_geocode(lk::invoke_t& cxt)
 		get_tz = cxt.arg(1).as_boolean();
 	}
 
-	double lat=NULL, lon=NULL, tz=NULL;
+	double lat = std::numeric_limits<double>::quiet_NaN();
+	double lon = std::numeric_limits<double>::quiet_NaN();
+	double tz = std::numeric_limits<double>::quiet_NaN();
 	bool ok = false;
 	cxt.result().empty_hash();
 
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -3814,7 +3814,7 @@ void fcall_geocode(lk::invoke_t& cxt)
 	// use GeoTools::GeocodeGoogle for non-NREL builds and set google_api_key in private.h
  	if (get_tz) {
 		ok = GeoTools::GeocodeDeveloper(cxt.arg(0).as_string(), &lat, &lon, &tz);
-		cxt.result().hash_item("tz").assign(tz); // converts NULL to 0
+		cxt.result().hash_item("tz").assign(tz);
 	}
 	else {
 		ok = GeoTools::GeocodeDeveloper(cxt.arg(0).as_string(), &lat, &lon);
--- a/src/invoke.cpp
+++ b/src/invoke.cpp
@@ -3794,19 +3794,32 @@ static bool copy_mat(lk::invoke_t &cxt, wxString sched_name, matrix_t<double> &m
 	return true;
 }
 
-void fcall_geocode(lk::invoke_t& cxt) 
+void fcall_geocode(lk::invoke_t& cxt)
 {
 	LK_DOC("geocode",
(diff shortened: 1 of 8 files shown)


src/invoke.cpp
```diff
@@ -3794,19 +3794,32 @@ static bool copy_mat(lk::invoke_t &cxt, wxString sched_name, matrix_t<double> &m
...
+	}
+
+	double lat=NULL, lon=NULL, tz=NULL;
+	bool ok = false;
 	cxt.result().empty_hash();
+
+	// use GeoTools::GeocodeGoogle for non-NREL builds and set google_api_key in private.h
+ 	if (get_tz) {
+		ok = GeoTools::GeocodeDeveloper(cxt.arg(0).as_string(), &lat, &lon, &tz);
+		cxt.result().hash_item("tz").assign(tz); // converts NULL to 0
```
sjanzou: Do we want to return 0 if the tz fails (no key)? tz=0 is valid for Greenwich.

No Azure key results
image

sjanzou (approved):
All tests do not crash with or without the Azure key.

The question is what to do for the tz return value when there is no key - should it stay 0?

Also, test 4 without a key has four messages to click through - all informative - I think that is a good workflow...but the invalid tz is set to NaN here which makes more sense than the script return value of 0 when there is a tz failure

image

image

src/invoke.cpp
cpaulgilman: TZ should be null or nan when it is not valid, thanks.

sjanzou (approved):
Looks good!

PR #2109, opened 2025-09-09
Ssc 195 day of week improvements

brtietz:
### Description

Add UI functionality for day of week behavior. This was implemented one of three places:

- Behind the meter models: load page
- Front of meter models: TOD factors page
- Merchant plant models: merchant plant description

The host developer model has the pages for both (1) and (2) so that required the implementation of UI variables. The core SSC variable is on the financial analysis page so that the compute modules have access to it. Open to ideas on a cleaner approach here.

The LCOE models and No financial configurations use the ?=0 in the vartable and continue to default to Monday (maintaning the previous behavior). We can open this up if users ask.

Fixes https://github.com/NREL/ssc/issues/195

### Corresponding branches and PRs:
Develop for lk and wex

Corresponding ssc PR https://github.com/NREL/ssc/pull/1352

### Unit Test Impact:

No expected changes to SAM tests. 

### Checklist
- [x] requires help revision and I added that label
- [x] adds, removes, modifies, or deletes variables in existing compute modules
- [ ] adds a new compute module
- [x] changes defaults
- [x] I've tagged this PR to a milestone

### Reminders- this section can be deleted
[Checking for PySAM Incompatible API Changes]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files).

[When do the PySAM files need to be regenerated?]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files#when-do-the-pysam-files-need-to-be-regenerated-via-export_config)

Issue #2105, opened 2025-08-26
Replace PVWatts-Battery with the detailed battery model


brtietz:
The PVWatts-Battery model is too limiting, especially concerning dispatch options. Replace the existing model with PVWatts-detailed battery as this has straightforward scaling and avoids a lot of problems with inverter sizing in detailed PV.

brtietz:
Completed with #2113

PR #2113, opened 2025-09-19
Sam 2105 use detailed batt w pvwatts

### LINKED ISSUE #753: Enable Lifetime Mode for PVWatts
We should consider enabling PVWatts in the SAM UI to run in lifetime mode.

brtietz:
### Description

Changes:

- Convert PVWatts to run in lifetime mode (UI and defaults)
- Swap to detailed battery model for the PVWatts-Battery Configuration
- Add FOM configurations for PVWatts-Battery
- Version upgrade script for the above

Two areas where I could use help:

1. The version upgrade script doesn't seem to be interpreting the battery chemistry and battery dispatch from the old pvwatts battery mode properly. These should be 0-1, and 0-2 respectively but I'm getting 3 for dispatch in my tests?
2. The progress bar goes from 0 - 200% for these cases.

Fixes https://github.com/NREL/sam/issues/753 and https://github.com/NREL/SAM/issues/2105

Just SAM, develop for other repos.

Note that https://github.com/NREL/SAM/pull/2109 will require some additional defaults changes in these new files, depending on merge order.

### Unit Test Impact:

Adds new configurations. Changes to lifetime mode mean defaults in test_results_win64 are expected.

### Checklist
- [X] requires help revision and I added that label
- [X] adds, removes, modifies, or deletes variables in existing compute modules
- [X] changes defaults

### Reminders- this section can be deleted
[Checking for PySAM Incompatible API Changes]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files).
- Adds new configurations, new files generated

[When do the PySAM files need to be regenerated?]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files#when-do-the-pysam-files-need-to-be-regenerated-via-export_config)


cpaulgilman (changes requested):
Please change the navigation tree label from "PV" to "PVWatts" to distinguish between Detailed PV and PVWatts models and for consistency with Hybrid configurations.

cpaulgilman:
@brtietz I made some changes to the Version Upgrade script to fix the problems you described and a couple of others. Please test and let me know if you notice other issues.

One unresolved item is that opening an old PVWatts-Battery file with lead acid battery does not correctly set the battery lifetime options. I think we want to load battery defaults for the chemistry and then apply the input from the old file, but not sure of the best way to do that.

brtietz:
@cpaulgilman Thanks for those fixes and feedback!

I pulled the appropriate code from the UI forms into the version upgrade script to fully handle the chemistry transition between lead acid flooded and Li-ion NMC. This seemed better than hardcoding the variables in cmod_battwatts, and trying to force the upgrade script and UI form to work together (which resulted in "on load" overwriting user values when I tried it). Does this seem reasonable to you?

I also found and fixed an error with the exclusive forms in the dispatch upgrades, and made the nav tree change you requested.

cpaulgilman (approved):
Thanks for these changes. Navigation menu looks good, and battery defaults are consistent with battery type.

PR #2121, opened 2025-10-02
Updates to scripting and reports for CPA macro functionality

sjanzou:
Changes

casewin - add ExportCasflowExcel
reports - fix multiple table header lines
script - add fcall_export_cashflow_excel
macro - include invoke.h
results - move Export from private to public

Easiest way to test is to 
1. copy the deploy/runtime/macros/Community Solar/Community Power Accelerator.lk macro over to this branch (but do not commit) 
2. copy the deploy/runtime/quickstart/CPA-CommunitySolar.samreport report template (but  do not commit) 
3. create a PVWatts/Community solar case
4. run the Community Power Accelerator macro
5. Check that the cash flow exports and the pdf report has a 2-line column header

### DIFF
--- a/src/casewin.cpp
+++ b/src/casewin.cpp
@@ -498,25 +498,36 @@ bool CaseWindow::RunSSCBaseCase(wxString& fn, bool silent, wxString* messages)
 
 bool CaseWindow::ExportCashflowExcel()
 {
-/*	// run base case automatically
-	if (!RunBaseCase())
-	{
-		wxMessageBox("Base case simulation did not succeed.  Please check your inputs before creating a report");
-		return;
-	}
-	*/
-	if (auto *vv= m_case->BaseCase().GetOutput("annual_energy")) {
+	if (m_case != NULL) {
+		if (m_case->GetFinancing() == "None") {
+			wxMessageBox("Case must have a financial model selected.");
+			return false;
+		}
+		else {
+			if (m_case->BaseCase().Ok()) {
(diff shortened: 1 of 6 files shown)


cpaulgilman (changes requested):
For LK function export_cashflow_excel():

1. Function returns Null instead of boolean:

image

image

2. For PVWatts / No Financial case, function opens empty Excel file. For configs with no financial model, I think it should return False and not open Excel.

3. The function runs a simulation, which means that in some cases (like the CPA macros) it runs a simulation when there are already results. It might be best to a) not run a simulation and return a message if there are no results, or b) add a flag parameter that determines whether to run a simulation.

4. In Linux, the function runs and returns Null. I think it should return False.

5. What is the "page name" input parameter? 

6. After opening Excel, the function displays the SAM Results window. I think that should be left to the script writer: If they want to display a page, they can use show_page() after calling the function.

sjanzou:
With commit https://github.com/NREL/SAM/pull/2121/commits/0a4b9c0c702272758f19840da3ac06cda0ca4cbc, the fcall_export_cashflow_excel performs as follows:
1. always returns a boolean
2. no case - returns 0 (false from script)
image
3. case with no financial model - returns 0 (false from script)
image
4. case not yet run  - returns 0 (false from script)
image
5. after base case run - returns 1 (true from script)
image

cpaulgilman (approved):
Thanks @sjanzou. I made some additional changes based on your updates. I think this looks good now.

PR #2141, opened 2025-12-12
Update Help and About Windows

cpaulgilman:
Goes with https://github.com/NREL/SAM-private/pull/139

This PR is in preparation for migrating to open source Help generation tools (currently in progress, see https://github.com/NREL/SAM/tree/help-to-sphinx).

In this PR:

* `HelpWin` class is used to display "About SAM" information only. It uses wxHtmlWindow for all platforms. Note that wxHtmlWindow can only display a subset of HTML, so About SAM text must use only basic HTML. Alternative wxWebView has full-featured HTML display capability, but requires special handling for Windows that we want to avoid: https://docs.wxwidgets.org/3.2/classwx_web_view.html.

* Help HTML files are displayed using `wxLaunchDefaultBrowser` on all platforms. This displays Help on computer's default web browser. 

### Pros and Cons

Pros:
* Implementation and behavior is consistent across Windows, Mac, Linux and for official and open source builds of SAM.
* Single implementation of `HelpWin` in main.cpp instead of different versions in main_add.h for SAM and SAM-private is easier to maintain.
* No platform-specific code to display Help.
* New Sphinx-generated Help should be displayed correctly on all platforms, assuming SAM is installed on computer with modern browser set as default.

Cons:
* Help opens HTML in new web browser tab/window each time user clicks Help button/F1 instead of displaying it in same window.
* Help display depends on computer's default browser setting that we can't control.

## To Test

I have tested on Windows 11, macOS Tahoe 26.1 on M1 chip, and Linux Mint 21.1 Cinnamon on Intel.

1. Start SAM.
2. On Welcome page, click **Help contents**. Help should open in web browser.
3. Click **About**. About SAM window should open displaying either official or open source information.
4. In About SAM window, click **Open NLR release notes...**. Web browser should open with SAM release notes.
5. Create a case and go to any input page.
6. Click **Help** or press F1 key. Help topic should display in Web browser. Note that for Mac, context help no longer works so Help displays Introduction topic. This will be fixed in separate PR with new Help implementation.
7. Click **About**. About SAM window should open.
8. In About SAM window, click **Open NLR release notes...**. Web browser should open with SAM release notes.


src/main.cpp
```diff
@@ -38,12 +38,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <wx/stc/stc.h>
 #include <fstream>
 
-#if defined(__WXMSW__)||defined(__WXOSX__)
-#include <wx/webview.h>
-#else
-#include <wx/html/htmlwin.h> // for linux - avoid webkitgtk dependencies
-#endif
+//#if defined(__WXMSW__)||defined(__WXOSX__)
```
sjanzou: Go ahead and remove theses since everything functions much better now!

src/main.cpp
```diff
@@ -2456,67 +2464,141 @@ wxArrayString SamApp::RecentFiles()
...
-		if ( ! context.IsEmpty() )
-			url = "file:///" + fn.GetFullPath( wxPATH_NATIVE ) + context + ".html";
-		wxLaunchDefaultBrowser( url );
-		return;
-#else
-		if ( ! context.IsEmpty() )
-			url += "?" + context + ".html";
+		CreateAboutHtml();
+
+#ifdef __WXMSW__
```
sjanzou, line 2469: Thanks for adding!

sjanzou (approved):
Everything functions well and looks great...

Not sure where name changes go (URLs seem beyond SAM) but some to the disclaimer text and the copyright notices will need to be updated somewhere and sometime

Image

PR #2185, opened 2026-05-06
pvwatts fom battery fixes

brtietz:
Working on another PR, I realized that the cashflow was not updated back in https://github.com/NatLabRockies/SAM/pull/2113

Clean that up as well as a few other FOM battery defaults


cpaulgilman (approved):
Thanks for catching and fixing.

PR #2189, opened 2026-05-11
Csp controller target net power

tyneises:
Add features corresponding to ssc pull request: https://github.com/NatLabRockies/ssc/pull/1396

Also adds estimate of actual solar multiple to target solar multiple reports difference on System Design page.

image


tyneises:
This also comments out CSP hybrids from the startup file that @mjprilliman added in the hybrid branch

tyneises:
@cpaulgilman What do you think about adding an option in the MSPT UI for this "beta" feature as shown above?

taylorbrown75 (approved):
Branch builds on my machine and successfully targets the net power output in the default case.

cpaulgilman:
> @cpaulgilman What do you think about adding an option in the MSPT UI for this "beta" feature as shown above?

@tyneises That approach looks good to me. It clearly communicates that the plant control target to net electrical output is in Beta, and the simulation output is what I expect based on the UI text description.

The System Design help topic will need to be updated.

mjprilliman (approved):
Hybrids seems to work correctly.

PR #2220, opened 2026-06-19
Updates to new ITC-depreciation workflow

janinefreeman:
Fixes typo in one of the option selections and adds units to various UI objects that are missing units.

@mjprilliman and @brtietz Is this a bug? I would expect ITC qualifying to be 0 based on the allocations that are set up? Or, if these calculated boxes are no longer relevant with the second radio button selection, they should be greyed out or disappear.
image


mjprilliman:
> Fixes typo in one of the option selections and adds units to various UI objects that are missing units.
> 
> @mjprilliman and @brtietz Is this a bug? I would expect ITC qualifying to be 0 based on the allocations that are set up? Or, if these calculated boxes are no longer relevant with the second radio button selection, they should be greyed out or disappear. image

I vote grey out when not in use. I'm not familiar enough with the ssc changes to know if we can zero out the values when not in use, zeroing out also loses people's selections if they go back to the first option. @brtietz would have to weigh in on that part.

janinefreeman:
More specifically, the equation for those calculated values could be conditionally calculated from the selections in the "depreciation class allocations and bonus depreciation" section, if that radio button option is selected, versus being calculated from the inputs above. I'm not suggesting to zero them out.

mjprilliman (approved):

PR #2233, opened 2026-06-29
New PV Module and Inverter Libraries for 2026

cpaulgilman:
Update module and inverter libraries with data from CEC workbooks from 6/23/2026.

This is the first use of Python with IPOPT for the module library.

Use Develop for other repos.

### Unit Test Impact:

Changes defaults for Detailed PV, PV Battery, and Photovoltaic Wind Battery Hybrid configurations.

### Checklist
- [ ] changes defaults


brtietz:
Version and date columns both say 2025 - should these be 2026?

image

mjprilliman (commented):
In the module list, the STC column is rounded down for every entry (ex. Aptos Solar default for PV None rounds from 530.748 W for actual STC Pmp to 530.0 in the library. This seems to only effect the library listing, not the calculated values. Values seem to round to .0 or .5

Image

Can the A_c column be cut off at 3 digits after the decimal? 

For inverters, cec_date looks better before as DD/MM/YYYY rather than the format including the time 

Image

cpaulgilman:
@mjprilliman Thanks for the review.

> In the module list, the STC column is rounded down for every entry (ex. Aptos Solar default for PV None rounds from 530.748 W for actual STC Pmp to 530.0 in the library. This seems to only effect the library listing, not the calculated values. Values seem to round to .0 or .5

The Python script reads STC data from the CEC workbook instead of calculating it from as Pmp =  Vmp * Imp as the old LK script did. The **Max power voltage (Vmp)** value shown on the page is calculated. The purpose of STC in the library is for sorting, so I think this is OK for now. We can revisit later.

> Can the A_c column be cut off at 3 digits after the decimal?

I manually fixed that for now. We'll need to modify the Python script to format numbers so they don't have so many decimal places.

> For inverters, cec_date looks better before as DD/MM/YYYY rather than the format including the time

I removed the time stamp.

cpaulgilman:
@brtietz 

> Version and date columns both say 2025 - should these be 2026?

Yes. Thanks for catching that.

brtietz (approved):
Looks good, thanks for updating!

Flagging that these changes might render some of the API autogen updates on https://github.com/NatLabRockies/SAM/pull/2228/changes out of date, so we should run that one last time after things are merged.

PR #2246, opened 2026-07-08
Update citation for System Advisor Model version

brtietz:
Merge citation updates to main, then pull through to patch & develop


cpaulgilman (approved):
Thanks for fixing.

PR #2255, opened 2026-07-21
Add additional libssc.so call to get alma linux build working

brtietz:
Add libssc.so to release build path for alma linux build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,7 @@ else()
 			    NAMES lk.a lk.lib
 			    PATHS $ENV{LK_LIB} $ENV{LKDIR}/build $ENV{LKDIR}/build/Release)
 	    find_library( SSC_LIB
-			    NAMES ssc.dylib ssc.lib ssc.so
+			    NAMES ssc.dylib ssc.lib ssc.so libssc.so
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -272,7 +272,7 @@ else()
 			    NAMES lk.a lk.lib
 			    PATHS $ENV{LK_LIB} $ENV{LKDIR}/build $ENV{LKDIR}/build/Release)
 	    find_library( SSC_LIB
-			    NAMES ssc.dylib ssc.lib ssc.so
+			    NAMES ssc.dylib ssc.lib ssc.so libssc.so
```
sjanzou, line 275: Not sure why you need libssc.so. Historically, we have used ssc.so for all releases and libssc.so is only generated  or looked for when SAMAPI_EXPORT is defined. Typical uses are pysam and HPC or pvwatts standalone builds of ssc.

sjanzou (commented):
In all of the releases to date, ssc.so and not libssc.so is used, e.g. in the 2026.7.3 release:

Image

Can you please  remove the libssc.so find library call and build again with SAMAPI_EXPORT=0?

Thanks!

CMakeLists.txt
brtietz: It looks like the key command is `--target ssc` as a part of the make command. I'll close this PR and think about how to make that more obvious in the instructions.

PR #2257, opened 2026-07-22
Pysam 8 doc

cpaulgilman:
These are revisions to PySAM autodocumentation to go with PySAM 8.0.0 release for NREL to NLR changes.


brtietz (approved):
Thanks for fixing these! Always more renaming to do!

sjanzou (approved):
@cpaulgilman , The only remaining "NREL" references I could find in this SAM branch are in geotools.cpp and main_add.h (the registry issue we reverted)

Image

The other repo's patch branches still have some NREL hanging around, e.g.:

Image

PR #2259, opened 2026-08-17
Fix Ubuntu runner differ for NLR_PySAM-8.0.0

sjanzou:
Ubuntu runner update to properly handle 2026.7.3 release with NLR_PySAM-8.0.0 release for the python differ step.


brtietz (approved):
Thanks for fixing!

PR #2265, opened 2026-09-16
Ready to be tested

sjanzou:
Goes with ssc pull request 1424


src/invoke.cpp
```diff
@@ -1969,6 +1969,47 @@ void fcall_var_exists(lk::invoke_t& cxt)
 		cxt.result().assign((double)0);
 }
 
+
+void fcall_var_exists_hybrid(lk::invoke_t& cxt)
```
mjprilliman: It doesn't look like this is being called anywhere?

